PR #13, opened 2024-01-10
Adds some envvars needed by telemetry to each pod definition.

decko:


deploy/clowdapp.yaml
```diff
@@ -518,6 +528,18 @@ parameters:
   - name: PULP_WORKER_MEMORY_LIMIT
     description: Limit of memory use by Worker pods
     value: "512Mi"
+  - name: TELEMETRY_ENABLED
+    description: Enables the application to send telemetry data.
+    value: True
```
dkliban, line 533: value: "true"

dkliban (approved):

PR #14, opened 2024-01-10
Add patch to fix content_unit fixture

mikedep333:
https://github.com/pulp/pulpcore/pull/4882


mikedep333:
Note: We did this on pulp-clowder-deployments also.

PR #30, opened 2024-01-11
Adds contentsources user and domain.admin role

dkliban:
This patch adds a job that will:

  - create a new role called 'domain.admin'
  - create a user 'contentsources'
  - assign the 'domain.admin' role to the 'contentsources' user

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -74,6 +74,11 @@ objects:
       # Generate admin password
       openssl rand -base64 32 | tr -d '\n' > /tmp/password
       kubectl create secret generic pulp-admin-password --from-file /tmp/password
+      rm /tmp/password
+
+      # Generate contentsources user password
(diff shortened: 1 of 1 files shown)


deploy/clowdapp.yaml
```diff
@@ -394,6 +399,34 @@ objects:
...
+            - name: CONTENT_SOURCES_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: pulp-content-sources-password
+                  key: password
+                  optional: false
+            - name: XDG_CONFIG_HOME
+              value: "/tmp"
+            - name: XDG_CACHE_HOME
+              value: "/tmp"
```
mikedep333, line 429: Nice, yeah. I had to set these vars in unit_test.sh for the pytest command.

mikedep333 (approved):
☁️

PR #59, opened 2024-03-12
always treat user and role creation as having completed

jlsherrill:
but let role-assignment properly fail

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -445,10 +445,10 @@ objects:
           args:
           - "-c"
           - |
-            pulp config create --base-url http://pulp-api-svc:24817 --api-root /api/pulp/ --username admin --password "$PULP_ADMIN_PASSWORD" --domain default && \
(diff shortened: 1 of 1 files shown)


deploy/clowdapp.yaml
```diff
@@ -445,10 +445,10 @@ objects:
           args:
           - "-c"
           - |
-            pulp config create --base-url http://pulp-api-svc:24817 --api-root /api/pulp/ --username admin --password "$PULP_ADMIN_PASSWORD" --domain default && \
-            pulp role create --name domain.admin --permission core.add_domain --permission core.add_headercontentguard --permission core.add_compositecontentguard --permission rpm.add_rpmrepository --permission rpm.add_rpmremote --permission rpm.add_rpmdistribution --permission rpm.add_rpmpublication && \
-            pulp user create --username contentsources --password "$CONTENT_SOURCES_PASSWORD" && \
-            pulp user role-assignment add --username contentsources --role domain.admin --object ""
```
jlsherrill, line 451: i wasn't 100% sure why these were&&'d together?

dkliban: because i don't know what i am doing.

dkliban (approved):

PR #63, opened 2024-04-04
Bump pulpcore and pulp_rpm

dkliban:
we can stop carrying the test fixture patch. it is available in pulpcore 3.50.2


mikedep333 (approved):
Thanks for the update!

PR #66, opened 2024-04-15
Use HEAD requests for livenessProbe of API

dkliban:


mikedep333 (approved):

decko:
> This doesn't work because a livenessProbe only has an httpGet config. No such thing as httpHead. A liveness probe can be a command inside the container or a GET request.

are you kidding me?!

PR #70, opened 2024-04-16
Do not create the .docker directory in the workspace

mikedep333:


dkliban (approved):

mikedep333:
@dkliban FYI: CI was failing for an external reason. My force push was of an amended commit with only a different timestamp.

PR #75, opened 2024-04-25
Update pulpcore to the 3.51 branch.

decko:


mikedep333 (approved):

dkliban (changes requested):
pulpcore 3.52.0 is now available. 3.52 contains the liveness API. Switching to 3.52 will require removing the patch we are carrying right now. 

Let's use a full version string so that each time we redeploy, the exact same version is installed.

dkliban (approved):

PR #79, opened 2024-05-06
Adds a ClowderJob to create insights-cicd resources

dkliban:
Creates a insights-cicd user and domain that belongs to that user. Creates a rubygems.org pull through cache.

PULP-59


bmbouter (approved):
This all looks right to me! Thanks for making this!

PR #80, opened 2024-05-06
Revert "Adds a ClowderJob to create insights-cicd resources (#79)"

dkliban:
This reverts commit db4cc314a3443f445c275257600f0da1bf718fd7.

This job needed to only run in production one time. Not needed in other environments.


bmbouter:
Let's revert this!

bmbouter (approved):

PR #81, opened 2024-05-10
Patch opentelemetry-python-contrib

dkliban:
Due to an issue https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2513

We can remove the patch once https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2514 is released.


bmbouter:
Thank you! 🙏

PR #85, opened 2024-05-10
Reduce backlog to 64 for api workers

dkliban:
This reduces how many clients can be in a waiting state to receive a response.


decko (approved):

dkliban:
closing. will try increasing number of gunicorn workers first.

PR #97, opened 2024-05-27
Add a new job to create the Edge credentials.

decko:


decko:
Let's close this for now and return to it when they decide do go to production

PR #104, opened 2024-06-06
Deploy API and Content on port 8000

dkliban:
This change is needed to take advantage of clowder managing the web services.

THe deployment now enables a public web service for the API and Content.  The otel-collector has a private web service enabled. The API and Content are configured to send telemetry to the otel collector on the new port.

The web services we deploy manually remain in place for now. They will be removed once references to them are changed to the auto generated services in app-interface.

This also removes ingress creation for from ephemeral. Clowder handles this also.

This also got rid of the readyz.py that was used as part of the readiness probe. The probe 
was changed to a GET request.


mikedep333 (approved):
☁️

PR #138, opened 2024-06-25
Update to use virtualenv so as to safely pass

mikedep333:
the RPM modified file check

PULP-108


dkliban (approved):
Thank you! This is great!

PR #183, opened 2024-07-04
Add a local plugin to deal with customs for pulp-service and avoid patches.

decko:


dkliban (changes requested):
1. Change the version to 0.1.0
2. Remove everything from the /docs directory
3. Remove everything from the /tests directory
4. Remove the empty tasks/publishing.py and tasks/synchronizing.py files.

dkliban (approved):

PR #225, opened 2024-08-08
Change the order of declared deployment.

decko:


lubosmj (commented):
Is this required to reduce the number of errors that we record because the collector is deployed after pulp? We should maybe add a dependency here, or not?

dkliban:
Closing this because this was made in another PR yesterday.

PR #227, opened 2024-08-12
Sums all the instances and change the legends for the graph

decko:


lubosmj (commented):
This removes the two graphs in question we talked about. But, is not a 2m step too small?

decko:
> This removes the two graphs in question we talked about. But, is not a 2m step too small?

The step is just a mark on the graph. Since it's not used in any calculation, I think it's Ok.

lubosmj (approved):

PR #228, opened 2024-08-13
Adds DomainBasedPermission class

dkliban:
The DomainOrg model stores the org_id associated with a Domain. This relationship is recorded using a post create hook for Domains.

The DomainBasedPermission class grants permission to users who's org_id matches the requested Domain's org_id.


pulp_service/pulp_service/app/authorization.py
```diff
@@ -0,0 +1,47 @@
...
+import json
+import jq
+from pulpcore.app import settings
+from pulp_service.app.models import DomainOrg
+from rest_framework.permissions import BasePermission
+
+
+org_id_var = ContextVar('org_id')
+
+class DomainBasedPermission(BasePermission):
```
lubosmj, line 13: Later, we can do something similar to what we have done in pulp_container: https://github.com/pulp/pulp_container/blob/aafa1c8bd01b6708df7aef5277b517a5533fb0e3/pulp_container/app/registry_api.py#L240-L246. I do not know how the deployment would look like but I guess we can have a setting that will allow us to use either this permission class or default ones.

pulp_service/pulp_service/app/authorization.py
```diff
@@ -0,0 +1,47 @@
...
+            if header_content:
+                header_decoded_content = b64decode(header_content)
+            else:
+                return False
+        except Base64DecodeError:
+            return False
+        # Get the Org ID from the header
+        try:
+            header_value = json.loads(header_decoded_content)
+            json_path = jq.compile(".identity.internal.org_id")
```
lubosmj, line 31: Can we move this to the module level so we will pre-compile the query just once?

pulp_service/pulp_service/app/models.py
```diff
@@ -0,0 +1,12 @@
+from django.db import models
+from pulpcore.plugin.models import Domain
+class DomainOrg(models.Model):
+    """
+    One-to-many relationship between org ids and Domains.
+    """
+
+    org_id = models.CharField(db_index=True)
+    domain = models.ForeignKey(Domain, related_name="domains", on_delete=models.CASCADE, db_index=True)
```
lubosmj, line 9: > A database index is automatically created on the ForeignKey. You can disable this by setting [db_index](https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.Field.db_index) to False. You may want to avoid the overhead of an index if you are creating a foreign key for consistency rather than joins, or if you will be creating an alternative index like a partial or multiple column index.

See https://docs.djangoproject.com/en/3.2/ref/models/fields/#foreignkey.

lubosmj: By the way, should not the relation be reversed (foreignkey defined on the Domain model)? I thought we wanted to ensure that each domain can be a member of exactly one organization. The same orgID should be then used to access multiple domains.

Domain.org => a single orgid attached to one domain.
org.domains.all() => multiple domains under the same orgid.

dkliban: Users should be allowed to create multiple domains.

dkliban: Adding a foreign key to an org would require us making a change in pulpcore. I don't think this makes sense for pulpcore.

pulp_service/pulp_service/app/authorization.py
dkliban: I'll move it. Thanks for the suggestion.

pulp_service/pulp_service/app/authorization.py
```diff
@@ -0,0 +1,52 @@
...
+import jq
+from pulpcore.app import settings
+from pulp_service.app.models import DomainOrg
+from rest_framework.permissions import BasePermission
+import logging
+
+
+_logger = logging.getLogger(__name__)
+org_id_var = ContextVar('org_id')
+org_id_json_path = jq.compile(".identity.internal.org_id")
```
decko, line 14: Should we make this configurable to make it work with cert and console.dot service accounts? Do they use the same path to their `org_id`s?

pulp_service/pulp_service/app/migrations/0001_initial.py
```diff
@@ -0,0 +1,27 @@
...
+        ('core', '0120_get_url_removal'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='DomainOrg',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('org_id', models.CharField(db_index=True)),
+                ('domain', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='domains', to='core.domain')),
```
decko, line 21: Do we want the `org_id` to be deleted when the domain is gone?

decko (approved):

PR #237, opened 2024-08-22
Consider users/groups when running DomainBasedPermission checks

lubosmj:


pulp_service/pulp_service/app/authorization.py
```diff
@@ -44,12 +46,28 @@ def has_permission(self, request, view):
...
+
+
+class DomainPermissionChecker:
+    def __init__(self, request, org_id):
+        self.user = request.user
+        self.domain_name = request.META['PATH_INFO'].removeprefix(settings.API_ROOT).split("/")[0]
+        self.org_id = org_id
+
+    def has_permissions(self):
+        return self.has_org_id_perms() or self.has_user_perms() or self.has_group_perms()
```
lubosmj, line 64: Does the order matter when checking for the user-domain permissions? Should we prioritize one over the other?

bmbouter: I suspect we want the lowest latency with user or group permissions (put those first) and then ord_id permissions last because that's likely a human doing it.

This I think is fine for now, but one future idea would be to produce one query that checks all 3 things at once and brings all the data across the cursor once instead of going back to the DB 3 times.

dkliban:
closing in favor of https://github.com/pulp/pulp-service/pull/239

PR #239, opened 2024-08-26
Consider users when running DomainBasedPermission checks

dkliban:


pulp_service/pulp_service/app/authorization.py
```diff
@@ -23,33 +25,50 @@ def has_permission(self, request, view):
         # Admins have all permissions
         if request.user.is_superuser:
             return True
+
+        domain_name = request.META['PATH_INFO'].removeprefix(settings.API_ROOT).split("/")[0]
```
bmbouter, line 29: isn't there a get_domain already? I'm totally unsure

pulp_service/pulp_service/app/authorization.py
```diff
@@ -23,33 +25,50 @@ def has_permission(self, request, view):
...
+            return True
+
+        # Check if user is creating a domain
+        if request.META['PATH_INFO'].endswith('/default/api/v3/domains/') and request.META['REQUEST_METHOD'] == 'POST':
+            action = "domain_create"
+        else:
+            action = "domain_operation"
+
         # Decode the identity header
+        header_present = False
```
bmbouter, line 43:
```suggestion
        RH_identity_header_present = False
```

pulp_service/pulp_service/app/signals.py
```diff
@@ -1,15 +1,23 @@
...
         org_id = org_id_var.get(None)
         # The default domain is created when migrations are run for the first time.
         # org_id is only available when REST API is used to create the domain.
         if org_id:
             DomainOrg.objects.create(domain=kwargs['instance'], org_id=org_id)
+
+        user_id = user_id_var.get(None)
+        if user_id:
+            user = get_user_model().objects.get(pk=user_id)
+            DomainOrg.objects.create(domain=kwargs['instance'], user=user)
```
bmbouter, line 23: It would be better to make this a **kwargs tyle call to have a single `DomainOrg.objects.create()` call in all cases.

pulp_service/pulp_service/app/models.py
```diff
@@ -1,12 +1,15 @@
...
-    org_id = models.CharField(db_index=True)
-    domain = models.ForeignKey(Domain, related_name="domains", on_delete=models.CASCADE, db_index=True)
-
-    class Meta:
-        unique_together = ("org_id", "domain")
+    org_id = models.CharField(null=True, db_index=True)
+    domain = models.ForeignKey(Domain, related_name="domains", on_delete=models.CASCADE)
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL, related_name="users", on_delete=models.CASCADE
+    )
```
bmbouter, line 15: I think these are indexed by default, but are we sure? Maybe just put `db_index=True` to be explicit (and sure)?

pulp_service/pulp_service/app/authorization.py
```diff
@@ -23,33 +25,50 @@ def has_permission(self, request, view):
...
             header_content = request.META.get('HTTP_X_RH_IDENTITY')
             if header_content:
                 header_decoded_content = b64decode(header_content)
+                header_present = True
                 # Temporarily log the header for debugging purposes
                 _logger.info(header_decoded_content)
             else:
-                return False
+                if action is not "domain_create":
+                    return False
```
bmbouter, line 53: It would be helpful to have a comment here calling out the case this catches.

pulp_service/pulp_service/app/authorization.py
```diff
@@ -23,33 +25,50 @@ def has_permission(self, request, view):
...
         except Base64DecodeError:
-            return False
+            if action is not "domain_create":
+                return False
```
bmbouter, line 56: It would be helpful to have a comment here calling out the case this catches.

bmbouter:
It would be interesting to see how this works with API endpoints that don't require auth. When 'False' is returned, does that deny access to the status API? I think in cases where we "don't know" why is arriving (no header, no user) maybe we should be returning None instead of False indicating we just don't know?

pulp_service/pulp_service/app/authorization.py
bmbouter, line 43: in this whole if-block I would move the "get header" cases to their own method or at least move them to their own if-block instead of mixing the "checking the header" part with the "is this request allowed" part.

pulp_service/pulp_service/app/authorization.py
```diff
@@ -23,33 +25,50 @@ def has_permission(self, request, view):
...
+
+        # The user that created the domain has permission to access that domain
+        if DomainOrg.objects.filter(domain__name=domain_name, user=user).exists():
+            return True
+
+        # Check if user is creating a domain
+        if request.META['PATH_INFO'].endswith('/default/api/v3/domains/') and request.META['REQUEST_METHOD'] == 'POST':
+            action = "domain_create"
+        else:
+            action = "domain_operation"
```
bmbouter, line 40: I would move this to its own method

dkliban:
> It would be interesting to see how this works with API endpoints that don't require auth. When 'False' is returned, does that deny access to the status API? I think in cases where we "don't know" why is arriving (no header, no user) maybe we should be returning None instead of False indicating we just don't know?

Views and ViewSets that don't require auth don't call into this code.

pulp_service/pulp_service/app/authorization.py
dkliban: yes there is ... i will try using that here.

pulp_service/pulp_service/app/authorization.py
```diff
@@ -23,33 +26,52 @@ def has_permission(self, request, view):
...
+            if decoded_header_content:
+                org_id_var.set(org_id)
+            user_id_var.set(request.user.pk)
+            return True
+        # User has permission if the org_id matches the domain's org_id
+        else:
+            return DomainOrg.objects.filter(domain=get_domain(), org_id=org_id).exists()
+
+    def get_user_action(self, request):
+        if request.META['PATH_INFO'].endswith('/default/api/v3/domains/') and request.META['REQUEST_METHOD'] == 'POST':
```
bmbouter, line 55: What about the case where it's a non-domain operation, e.g. listing a worker? I think that would mistakenly be marked here as `domain_operation`?

pulp_service/pulp_service/app/signals.py
```diff
@@ -1,15 +1,17 @@
...
 def post_create_domain(sender, **kwargs):
     if kwargs['created']:
-        from pulp_service.app.authorization import org_id_var
+        from pulp_service.app.authorization import org_id_var, user_id_var
         org_id = org_id_var.get(None)
-        # The default domain is created when migrations are run for the first time.
-        # org_id is only available when REST API is used to create the domain.
-        if org_id:
-            DomainOrg.objects.create(domain=kwargs['instance'], org_id=org_id)
+        user_id = user_id_var.get(None)
```
bmbouter, line 16: context_vars do support a default=None option which maybe would move these to the top instead?

pulp_service/pulp_service/app/authorization.py
```diff
@@ -23,33 +26,52 @@ def has_permission(self, request, view):
         # Admins have all permissions
         if request.user.is_superuser:
             return True
-        # Decode the identity header
+        user = request.user
+
+        # The user that created the domain has permission to access that domain
+        if DomainOrg.objects.filter(domain=get_domain(), user=user).exists():
```
bmbouter, line 32: I agree with what you said, it would be better to combine this with the `filter` further down. When combining use a Q() filter also so that it only goes to the DB once instead of two `filter()` calls combined with a Python `or`.

pulp_service/pulp_service/app/models.py
lubosmj: > A database index is automatically created on the ForeignKey. You can disable this by setting [db_index](https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.Field.db_index) to False. You may want to avoid the overhead of an index if you are creating a foreign key for consistency rather than joins, or if you will be creating an alternative index like a partial or multiple column index.

See https://docs.djangoproject.com/en/3.2/ref/models/fields/#foreignkey.

pulp_service/pulp_service/app/authorization.py
dkliban: Yes, it is a domain operation. You are listing workers within the default domain. No one except the admin has the right to use the default domain. We can give certain users access to the default domain if we decide to.

pulp_service/pulp_service/app/signals.py
bmbouter: I meant something like this:  https://github.com/pulp/pulpcore/blob/main/pulpcore/app/util.py#L450 I have no idea if this is good or not.

pulp_service/pulp_service/app/authorization.py
```diff
@@ -23,33 +26,49 @@ def has_permission(self, request, view):
...
+        decoded_header_content = self.get_decoded_identity_header(request)
+
+        # Get the Org ID from the Red Hat Identity header
+        org_id = self.get_org_id(decoded_header_content)
+
+        # Anyone can create a domain
+        if action == "domain_create":
+            if decoded_header_content:
+                org_id_var.set(org_id)
+            user_id_var.set(request.user.pk)
```
bmbouter, line 44: If there is no user set at all I think it would be good to `return False` here.

PR #249, opened 2024-09-12
Adds django-admin UI

dkliban:


lubosmj (commented):
This change seems work locally (with the merge directive). I can manage only groups and users. Is this what we wanted to achieve?

image

dkliban:
@lubosmj Take a look at my latest changes. The goal is to manage DomainOrg table.

lubosmj (approved):

PR #257, opened 2024-09-17
deploy: Add missing cpu limit and request

jrusz:
The job create-settings-and-ingress needs a cpu limit and request as well.

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -549,8 +549,10 @@ objects:
           resources:
             requests:
               memory: 192Mi
+              cpu: 250m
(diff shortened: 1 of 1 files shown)


jrusz:
Hi @dkliban can you take a look at this please? Bonfire fails to deploy this without the cpu limits, this is the related change https://github.com/RedHatInsights/bonfire/pull/421 ... There needs to be at least something set. I've just put those limits at random, so if you think of a better value please let me know and I'll update it.

dkliban (approved):

dkliban:
@jrusz thanks for the PR!

PR #259, opened 2024-09-17
Changes the order of Auth classes.

decko:


deploy/clowdapp.yaml
```diff
@@ -893,10 +893,10 @@ parameters:
     value: pulp-stage-default-domain-s3
   - name: PULP_AUTHENTICATION_BACKENDS
     description: The AUTHENTICATION_BACKENDS to be used by Pulp.
-    value: "@merge django.contrib.auth.backends.RemoteUserBackend" 
+    value: "django.contrib.auth.backends.RemoteUserBackend,django.contrib.auth.backends.ModelBackend,pulpcore.backends.ObjectRolePermissionBackend"
```
dkliban, line 896: This needs to be inside [ ] to be a list. Pulp API fails to start with this string in the settings.

decko: Done

deploy/clowdapp.yaml
```diff
@@ -899,10 +899,10 @@ parameters:
     value: pulp-stage-default-domain-s3
   - name: PULP_AUTHENTICATION_BACKENDS
     description: The AUTHENTICATION_BACKENDS to be used by Pulp.
-    value: "@merge django.contrib.auth.backends.RemoteUserBackend" 
+    value: "@merge django.contrib.auth.backends.RemoteUserBackend"
   - name: PULP_REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES
     description: The defaul authentication classes for Pulp.
-    value: "@merge pulpcore.app.authentication.JSONHeaderRemoteAuthentication"
+    value: "['pulpcore.app.authentication.JSONHeaderRemoteAuthentication','rest_framework.authentication.BasicAuthentication','rest_framework.authentication.SessionAuthentication']"
```
gerrod3, line 905: Is session auth (cookie auth) used at all for services? I would recommend putting it before BasicAuth if it is.

decko: I believe it's gonna be used with the django-admin that we enabled on services. Good catch.

dkliban: let's also add pulp_service.app.authentication.RHServiceAccountCertAuthentication to this list here. and if you make it first in the list, it will return the proper 401 (with the other change you already have in this PR).

pulp_service/pulp_service/app/authentication.py
```diff
@@ -5,3 +5,6 @@ class RHServiceAccountCertAuthentication(JSONHeaderRemoteAuthentication):
 
     header = "HTTP_X_RH_IDENTITY"
     jq_filter = ".identity.x509.subject_dn"
+
+    def authenticate_header(self, request):
+        return self.__name__
```
dkliban, line 10: This produce a run time error: AttributeError: 'RHServiceAccountCertAuthentication' object has no attribute '__name__'

Can you just change this to "Bearer" string?

dkliban (approved):
Thank you!

PR #273, opened 2024-09-23
Increase Pulp Content replicas to 10.

decko:


deploy/clowdapp.yaml
```diff
@@ -826,7 +826,7 @@ parameters:
     value: "1"
   - name: PULP_CONTENT_REPLICAS
     description: Number of pulp content replicas
-    value: "1"
+    value: "10"
```
dkliban, line 829: let's leave it at 1 here. the change we are looking for is in app-interface's deploy.yaml for Pulp.

PR #284, opened 2024-09-30
Enable Domain admin through django-admin.

decko:


pulp_service/pulp_service/app/admin.py
```diff
@@ -14,7 +16,14 @@ class DomainOrgAdmin(admin.ModelAdmin):
     list_editable = ["user", "org_id"]
     list_filter = ["user", "org_id"]
 
+
+class DomainAdmin(admin.ModelAdmin):
+    list_display = ["name", "description", "storage_class"]
+    list_display = ["name", "description", "storage_class"]
```
dkliban, line 22: line 22 is not needed.

decko: It was suppose to be a list_filter statement

dkliban (approved):

PR #288, opened 2024-10-07
Test artifact size header patch

lubosmj:


lubosmj:
This change resolved the issue: https://github.com/pulp/pulpcore/pull/5880. I can see the X-PULP-ARTIFACT-SIZE header after applying the patch.

PR #289, opened 2024-10-07
Patch the resource name and record the worker name

lubosmj:


lubosmj:
A copy of https://github.com/pulp/pulpcore/pull/5865. Needed before we release the official image.

decko (approved):

PR #292, opened 2024-10-08
Test a patch that add the worker name to all metric data points

decko:


bmbouter (approved):
I'm excited to see this!

PR #294, opened 2024-10-15
Adds middleware that generates cProfile profile for a request

dkliban:
The request needs to have a X-PROFILE-REQUEST header for the middleware to generate a profile. The middleware logs a URL that can be used to download the profile.

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -310,6 +310,8 @@ objects:
             value: ${PULP_CSRF_TRUSTED_ORIGINS}
           - name: PULP_STATIC_URL
             value: ${PULP_STATIC_URL}
+          - name: PULP_MIDDLEWARE
+            value: ${PULP_MIDDLEWARE}
 
     - name: content
       replicas: ${{PULP_CONTENT_REPLICAS}}
@@ -826,3 +828,6 @@ parameters:
   - name: PULP_STATIC_URL
     description: The path to the static assets needed by some Django apps
     value: "/api/pulp/assets/"
+  - name: PULP_MIDDLEWARE
+    description: List of Middleware
+    value: "@merge pulp_service.app.middleware.ProfilerMiddleware"
--- a/pulp_service/pulp_service/app/middleware.py
+++ b/pulp_service/pulp_service/app/middleware.py
@@ -1,0 +1,57 @@
+from contextlib import suppress
+import cProfile
+import logging
+import marshal
+import tempfile
+
+from django.db import IntegrityError
+from django.utils.deprecation import MiddlewareMixin
+
+from pulpcore.app.models import Artifact
+from pulpcore.app.util import get_artifact_url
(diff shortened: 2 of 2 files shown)


pulp_service/pulp_service/app/middleware.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    PROFILER_REQUEST_ATTR_NAME = '_django_cprofile_middleware_profiler'
+
+    def can(self, request):
+        if 'HTTP_X_PROFILE_REQUEST' in request.META:
+            return True
+        return False
+    def process_view(self, request, callback, callback_args, callback_kwargs):
+        if self.can(request):
+            profiler = cProfile.Profile()
```
bmbouter, line 31: We have this dependency already yes?

deploy/clowdapp.yaml
```diff
@@ -826,3 +828,6 @@ parameters:
   - name: PULP_STATIC_URL
     description: The path to the static assets needed by some Django apps
     value: "/api/pulp/assets/"
+  - name: PULP_MIDDLEWARE
+    description: List of Middleware
+    value: "@merge pulp_service.app.middleware.ProfilerMiddleware"
```
bmbouter, line 833: Are we sure this will be earliest in the middleware?

pulp_service/pulp_service/app/middleware.py
```diff
@@ -0,0 +1,57 @@
...
+    Simple profile middleware to profile django views. To run it, add
+    x-profile-request header to the request.
+
+    This is adapted from an example found here:
+    https://github.com/omarish/django-cprofile-middleware/blob/master/django_cprofile_middleware/middleware.py
+    """
+    PROFILER_REQUEST_ATTR_NAME = '_django_cprofile_middleware_profiler'
+
+    def can(self, request):
+        if 'HTTP_X_PROFILE_REQUEST' in request.META:
```
bmbouter, line 26: So the first draft is entirely yes/no based on the presence of the header? I'm ok with that if so.

pulp_service/pulp_service/app/middleware.py
dkliban: it's part of the standard library https://docs.python.org/3.8/library/profile.html

deploy/clowdapp.yaml
dkliban: It will be the last middleware in the list.

pulp_service/pulp_service/app/middleware.py
dkliban: that is correct

deploy/clowdapp.yaml
bmbouter: That's better than not in the list! Is there a way to make it first? I'm interested in investigating the basic auth middleware overhead for example. If not I'm ok with trying this as-is also.

PR #300, opened 2024-10-17
Add the file extension to the served artifact size metric

lubosmj:


lubosmj:
```
http :5001/pulp/content/default/test/3.iso
```
image

decko (approved):
I believe we need to test with others FQDN, but the solution seems promising. LTTM

PR #304, opened 2024-10-22
Add http_target to the pulp-api's server duration metric

lubosmj:


lubosmj:
image

My change should get us to this state ^, see below:

image


Note that this records the full domain name. Are we fine with it?

bmbouter:
I like that the full domain name is in there. Yes it's a large URL space, but it's really useful I think business-wise.

Anywhere you have a variable like `UUID` I think it should be represented like a path_parameter in openapi spec, e.g. `/pulp/default/api/v3/tasks/{uuid}/`. This uses lowercase names for the variable and puts it in curly braces.

lubosmj:
image

lubosmj:
We decided not to show the whole domain name; instead, we are going to match the previous behaviour.

lubosmj:
@bmbouter, how do you like it now?
image

bmbouter:
I think it's good and we should try it! Can you merge and deploy?

bmbouter (approved):

PR #307, opened 2024-10-23
Unquote paths in the http_target attribute

lubosmj:


bmbouter (approved):
I'd like to see this deployed.

PR #313, opened 2024-10-25
Add a custom metrics middleware for pulp-api

lubosmj:


lubosmj:
image

bmbouter:
This looks amazing!

pulp_service/pulp_service/app/middleware.py
```diff
@@ -55,3 +66,66 @@ def process_response(self, request, response):
...
+        exporter = OTLPMetricExporter()
+        reader = PeriodicExportingMetricReader(exporter)
+        resource = Resource(attributes={"service.name": "pulp-api"})
+        provider = MeterProvider(metric_readers=[reader], resource=resource)
+
+        set_meter_provider(provider)
+        meter = metrics.get_meter("pulp.custom_metrics")
+
+        self.request_duration_histogram = meter.create_histogram(
+            name="http.server.custom.request_duration",
```
bmbouter, line 82: Can we put the process name in here somehow so that the django custom histogram and the aiohttp custom histogram (not yet created) will no longer colide... ?

bmbouter: Because right now all of our filtering needs to filter one process or the other, it would be nicer for the graphs just to use "the correct metric" when selecting the data source

bmbouter: Maybe `http.api.request_duration` and then `http.content.request_duration`?
Or `http.server.custom.request_duration` vs `aiohttp.server.custom.request_duration`
I keep on thinking we should not have the word "custom" in it, they're just the Pulp metrics.
Naming is hard.

bmbouter (approved):
I'm LGTMing in case you want to merge as is and fixup later.

PR #317, opened 2024-10-28
Add a custom metrics middleware for pulp-content

lubosmj:
In this commit, the `pulp_http_server_custom_request_duration_milliseconds_*` metric was split into two `pulp_content_request_duration_milliseconds_*` and `pulp_api_request_duration_milliseconds_*` metrics.

Even though these metrics are self-explanatory, it is still recommended to specify "service.name". Otherwise, "exported_job" on the collector's side will be set to "unknown_service".


bmbouter (approved):
Thank you!

PR #320, opened 2024-10-30
Fix the range for time series in the dashboard

lubosmj:
Inspired by
https://hoelz.ro/blog/use-caution-when-using-rate_interval-along-with-increase.


lubosmj:
Before:
Screenshot from 2024-10-30 17-17-20

After:
Screenshot from 2024-10-30 17-17-16

decko (approved):

PR #322, opened 2024-10-30
Extent Update Support ContentGuard

decko:
BEFORE MERGING THIS, WE NEED:

[x] Add the key to our Staging Vault
[ ] Add the key to our Prod Vault
[ ] Make the key available through app-interface for staging
[ ] Make the key available through app-interface for production


pulp_service/pulp_service/app/settings.py
```diff
@@ -4,3 +4,7 @@
 .. _Plugin Writer's Guide:
     https://docs.pulpproject.org/pulpcore/plugins/plugin-writer/index.html
 """
+
+RH_IDENTITY_HEADER = "HTTP_X_RH_IDENTITY"
+SUBSCRIPTION_API_URL = "https://subscription.stage.api.redhat.com/svcrest/subscription/v5/featureStatus?accountId=5894300&features=RHEL-OS-x86_64&features=RHEL-EUS-x86_64"
```
bmbouter, line 9:
```suggestion
SUBSCRIPTION_API_URL = "https://subscription.stage.api.redhat.com/svcrest/subscription/v5/featureStatus"
```

You append this later I think right?

pulp_service/pulp_service/app/settings.py
```diff
@@ -4,3 +4,7 @@
...
+SUBSCRIPTION_CERT_PATH = "/var/certs/pulp_nonprod_cert.pem"
```
bmbouter, line 10: I think the key (and probably the cert itself too) to be in the vault and this should read from the vault. I'm not 100% sure how that works, but I think that's the idea. Also we'll have one cert/key for staging and a different one for prod.

pulp_service/pulp_service/app/viewsets.py
```diff
@@ -0,0 +1,74 @@
...
+    LOCKED_ROLES = {
+        "core.featurecontentguard_creator": ["core.add_featurecontentguard"],
+        "core.featurecontentguard_owner": [
+            "core.view_featurecontentguard",
+            "core.change_featurecontentguard",
+            "core.delete_featurecontentguard",
+            "core.manage_roles_featurecontentguard",
+        ],
+        "core.featurecontentguard_viewer": ["core.view_featurecontentguard"],
+    }
```
bmbouter, line 74: If this PR was going against pulpcore this would be right, but I don't think we should include any of this since we're not shipping it in upstream pulpcore. Our installation has all of this stuff disabled and wouldn't use any of it.

pulp_service/pulp_service/app/viewsets.py
```diff
@@ -0,0 +1,74 @@
...
+from pulp_service.app.models import FeatureContentGuard
+from pulp_service.app.serializers import FeatureContentGuardSerializer
+
+
+class FeatureContentGuardViewSet(ContentGuardViewSet, RolesMixin):
+    """
+    Content guard to protect the content guarded by Subscription Features.
+    """
+
+    endpoint_name = "feature"
```
bmbouter, line 12: Can you give me a URL example of interacting with this viewset? I think it's right, but I just was wanting an example.

bmbouter:
We can ask @dkliban when he's back about the options to test in ephemeral. It would be great to do against the "live" service, but I don't know enough about it.

pulp_service/pulp_service/app/settings.py
decko: There's a `volumeMount` on the deployment file to deal with that. Sadly the `requests` lib can't read a cert from an envvar, so I had to mount the certificate as a file in the container and then point `requests` to use that cert.

bmbouter: Mmm yeah I see the `volumeMount`, but where does the .pem file actually live?

Also a few years ago we tried to move away from requests as our library in favor of aiohttp.client. For example, AIUI we don't actually declare a dependency on requests (at least pulpcore doesn't). Here's an example of a client request using in memory certs with aiohttp.client.

```
import ssl
import aiohttp
import asyncio

# In-memory certificate and key (these should be in PEM format)
cert_data = b"""-----BEGIN CERTIFICATE-----\n...\n-----END CERTIFICATE-----"""
key_data = b"""-----BEGIN PRIVATE KEY-----\n...\n-----END PRIVATE KEY-----"""

# Create an SSL context
ssl_context = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
ssl_context.load_cert_chain(certfile=cert_data, keyfile=key_data)

# Function to make a request
async def fetch_with_cert():
    async with aiohttp.ClientSession() as session:
        async with session.get("https://example.com", ssl=ssl_context) as response:
            print(await response.text())

# Run the request in an event loop
asyncio.run(fetch_with_cert())
```

What do you think about all this?

pulp_service/pulp_service/app/viewsets.py
decko: I'm using like this:
```
echo -n '{
"name": "test-feature-contentguard",
"description": ".",
"header_name": "x-rh-identity",
"jq_filter": ".agent_id",
"features": [
"RHEL-EUS-x86_64",
"RHEL-OS-x86_64"
]
}' | http POST :5001/pulp/another_domain/api/v3/contentguards/service/feature/

dkliban (approved):

PR #354, opened 2024-11-21
Update grafana dashboard

git-hyagi:
* re-organize latency and rps panels by pulpcore pod types
* modify request rate queries (kept the old ones disabled to make it easier to roll back or compare the outputs)
* decreased the dashboard default "time range" to avoid error trying to pull a lot of metrics at once and the graphs "crashing"
* modified the "graph tooltip" to "shared tooltip"
* modified request rate panels legend to table and added the max/mean/last counters


git-hyagi:
new distribution
Screenshot from 2024-11-21 16-04-36

bmbouter (approved):
Let's do this!

dkliban (approved):

PR #360, opened 2024-11-25
[PULP-213] Add PULP_USE_UVLOOP envvar to activate uvloop in the content-app

pedro-psb:
Experiement with `uvloop`: https://github.com/pulp/pulpcore/issues/6021

Setting `PULP_USE_UVLOOP` to anything will enable it, while not having is set uses the regular worker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -143,6 +143,9 @@ RUN patch -p1 -d /usr/local/lib/pulp/lib/python${PYTHON_VERSION}/site-packages <
 COPY images/assets/patches/0019-Adds-LabelsMixin-to-ReadOnlyContentViewSet.patch /tmp/
 RUN patch -p1 -d /usr/local/lib/pulp/lib/python${PYTHON_VERSION}/site-packages < /tmp/0019-Adds-LabelsMixin-to-ReadOnlyContentViewSet.patch
 
+COPY images/assets/patches/0020-Add-uvloop-as-the-the-Gunicorn-worker-eventloop.patch /tmp/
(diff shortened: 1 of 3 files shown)


pedro-psb:
I dont know how to interpret/debug the konflux CI failure.

decko:
> I dont know how to interpret/debug the konflux CI failure.

It's Ok. It will keep failing and yet the build and deploy happens without issues.

images/assets/patches/0013-Add-uvloop-as-the-the-Gunicorn-worker-eventloop.patch
```diff
@@ -0,0 +1,46 @@
...
+   "tablib<3.6.0",
+   "url-normalize>=1.4.3,<=1.4.3",
+   "uuid6>=2023.5.2,<=2024.7.10",
++  "uvloop>=0.21.0,<0.22",
+   "whitenoise>=5.0,<6.9.0",
+   "yarl>=1.8,<1.18.1",
+ ]
+-- 
+2.47.0
+
```
bmbouter, line 46: Would you be ok if this part of the diff is pulled out since we're declaring the dependency directly below?

pedro-psb: Yes. In fact I was supposed to ask where was the right place at the time but I forget about it. Thanks for the input

bmbouter (changes requested):
@pedro-psb I'll lgtm and merge when you push next. Ty!

bmbouter:
@pedro-psb I should have told you this earlier (sorry) but you also need to copy the patch to the container build and run it. See this example: https://github.com/pulp/pulp-service/pull/442/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R143

PR #420, opened 2025-03-06
Content scan poc

git-hyagi:


pulp_service/pulp_service/app/viewsets.py
```diff
@@ -138,3 +149,42 @@ def get_queryset(self):
     @classmethod
     def view_name(cls):
         return "admintasks"
+
+
+class ContentScan(APIView):
+    authentication_classes = []
+    permission_classes = []
+
+    def get(self, request=None, prn=None):
```
bmbouter, line 158: From a usability perspective I think the argument here should be `repo_version` instead of `prn` although the value should be in the format `prn`. From an openAPI perspective the way this is currently it's polymorphic in the sense a PRN can refer to lots of different object types in Pulp. To keep things simple I think we should have this refer to a `repo_version` and then this viewset should provide validation that the prn is indeed referring to a repository version.

pulp_service/pulp_service/app/viewsets.py
```diff
@@ -138,3 +141,27 @@ def get_queryset(self):
...
+class ContentScan(APIView):
+    authentication_classes = []
+    permission_classes = []
+
+    def post(self, request=None):
+        serialized_data = ContentScanSerializer(data=request.data)
+        serialized_data.is_valid(raise_exception=True)
+        repo_version_pk = serialized_data.data["repo_version"]
+        task = dispatch(check_content, kwargs={"repo_version_pk": repo_version_pk})
+        return Response(task.pk)
```
dkliban, line 155: we should return an OperationPostponedResponse here similar to https://github.com/pulp/pulpcore/blob/main/pulpcore/app/viewsets/reclaim.py#L53

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -0,0 +1,105 @@
...
+    Makes a request to the osv.dev API and store the results in ArtifactVulnerability model
+    """
+    async with aiohttp.ClientSession() as session:
+        for package in packages:
+            osv_data = json.dumps(package["osv_data"])
+            async with session.post(url=OSV_QUERY_URL, data=osv_data) as response:
+                response_body = await response.text()
+                json_body = json.loads(response_body)
+                if json_body.get("vulns"):
+                    await sync_to_async(ArtifactVulnerability.objects.get_or_create)(
```
dkliban, line 45: I envisioned a single report per scan. The JSON would contain a key for each package. and the value for that key would be the scan results for that package.

pulp_service/pulp_service/app/viewsets.py
```diff
@@ -138,3 +142,27 @@ def get_queryset(self):
...
+        return OperationPostponedResponse(task, request)
+
+
+class Vulnerabilities(APIView):
+    def get(self, request, uuid=None):
+        queryset = ArtifactVulnerability.objects.all()
+        serializer = ArtifactVulnerabilitySerializer(queryset, many=True)
+        return Response(serializer.data)
+
+    def post(self,request):
```
bmbouter, line 165:
```suggestion
    def post(self, request):
```

bmbouter: The POST method in HTTP is used to send data to a server to create or modify a resource. Here it's being used to retrieve data. Here's what I recommend.

First let's talk about the urls. I'd make the design like this:

```
path("api/pulp/vulnerabilities/", Vulnerabilities.as_view()),
path("api/pulp/vulnerabilities/<uuid:uuid>/", VulnerabilitiesDetail.as_view()),
```

Have Vulnerabilities take GET (the list method you already wrote) and POST (the one that dispatches the task and takes the `repo_version` as an argument). It's common for GET to list and POST on that same endpoint to create a "new one".

Have VulnerabilitiesDetail implement the GET method which returns a serialized single Vulnerability report. This is what you've got implemented currently as Vulnerabilities.post().

This improves the design in 2 ways. (1) the HTTP method types are aligned with their standards based definitions of use (GET makes no server changes and just returns data and POST modifies or creates data). Also (2) it makes the API RESTful in the sense that there is no longer the verb, i.e. `scan` in the URL structure. Instead the user "creates a Vulnerability Report" using POST.

bmbouter: I would probably also change the path to be `api/pulp/vuln_report` instead of `api/pulp/vulnerabilities`. It's really a report, not a vulnerabilitiy itself. I think vulnerability_report would be too long so I'm just shortening it to vuln_report.

pulp_service/pulp_service/app/models.py
```diff
@@ -164,3 +165,12 @@ class Meta:
                 "Can manage role assignments on Feature ContentGuard",
             ),
         )
+
+
+class ArtifactVulnerability(models.Model):
```
bmbouter, line 170:
```suggestion
class VulnerabilityReport(models.Model):
```

pulp_service/pulp_service/app/models.py
```diff
@@ -164,3 +165,12 @@ class Meta:
...
+    """
+    Model used in content vulnerability scan
```
bmbouter, line 172:
```suggestion
    Model used in vulnerability report.
```

Similar replacement of `scan` with `report` throughout the PR.

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -0,0 +1,111 @@
+import aiohttp
+import logging
+import json
+
+from asgiref.sync import sync_to_async
+
+from pulpcore.app.models import Content, RepositoryVersion
+from pulp_rpm.app.models.package import Package as RPMPackage
```
bmbouter, line 8: you don't want this one here until we do the RPM work.

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -0,0 +1,111 @@
+import aiohttp
+import logging
+import json
+
+from asgiref.sync import sync_to_async
+
+from pulpcore.app.models import Content, RepositoryVersion
```
bmbouter, line 7: It's not safe to import directly from pulpcore.app. Any plugin importing from pulpcore should only import from `pulpcore.plugin` aka the "Plugin API" which is guaranteed stable except for "plugin API breaking change releases", e.g. 3.55, 3.70, 3.85, etc. You'll find all of these objects available in `pulpcore.plugin` See [here](https://github.com/pulp/pulpcore/blob/main/pulpcore/plugin/models/__init__.py).

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -0,0 +1,111 @@
...
+        vulns=scanned_packages,
+    )
+
+
+def _get_content_from_repo(repo_version_pk: str) -> list[any]:
+    """
+    Return the list of contents in the repository_version
+    """
+    repo_version = RepositoryVersion.objects.get(pk=repo_version_pk)
+    return Content.objects.filter(pk__in=repo_version.content)
```
bmbouter, line 61: I think you want to defer to the facilities pulpcore provides. Look at what you can get from the `RepositoryVersion` object. You could [list all the content units](https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/repository.py#L851-L852). The "downcast" there with `cast()` transforms the model object type from `Content` to for example FileContent or whatever the actual content type is. That may or may not be helpful to you.

I recommend iterating through all of the content handling the "supported types", e.g. Npm and Python initially and just ignoring the other types. It's efficient that way because you're just iterating once. Note as an FYI, you could also filter by type [like this](https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/repository.py#L832), but you'd be making N queries that way where N are the number of content ypes you're searching for, which would probably not be very efficient.

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    if not content_types:
+        return
+
+    package_list = []
+    for content in content_types:
+        content = content.cast()
+        ecosystem = _identify_package_ecosystem(content)
+        if not ecosystem:
+            break
```
bmbouter, line 77: Oh I see you're doing here exactly what I was recommending in the comment a few lines up! :+1:

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -0,0 +1,111 @@
...
+
+async def _identify_rh_cpe():
+    async with aiohttp.ClientSession() as session:
+        async with session.post(url=RH_REPO_TO_CPE_URL) as response:
+            response_body = await response.text()
+            json_body = json.loads(response_body)
+            global cache_rh_cpe
+            cache_rh_cpe = json_body["data"]
+            _logger.info(f"RESPONSE JSON: {json.dumps(cache_rh_cpe, indent=2)}")
+            # return json_body
```
bmbouter, line 111: this look unused so I'd remove it.

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -0,0 +1,111 @@
...
+        package_list.append(
+            {
+                "osv_data": {
+                    "package": {"name": content.name, "ecosystem": ecosystem},
+                    "version": content.version,
+                },
+                "content": content,
+            }
+        )
+    return package_list
```
bmbouter, line 87: You're building a potentially large in-memory data structure. I can see why because the Django ORM has to be run syncronously yet you want to handle the items in an asynchronous coroute so they aren't compatible.

Consider this example:

```
# Synchronous Django ORM code
def fetch_objects():
    return MyModel.objects.all()  # Example queryset

# Coroutine to process items one-by-one
async def process_items():
    # Run the synchronous Django ORM code in a thread
    queryset = await asyncio.to_thread(fetch_objects)

    # Iterate over the queryset and process items one-by-one
    for item in queryset:
        # Simulate async processing
        await asyncio.sleep(0.1)  # Replace with your async logic
        print(f"Processing item: {item}")
```

This is pretty nice because it lets your ORM code all live in `fetch_objects` and they can be handled one-by-one in an async model. This simplifies the whole matter by letting you handle them "directly, one-by-one" in the async coroutine instead of having to build a big data structure and then later loop through it in your coroutine. Try porting your code to this for fun.

pulp_service/pulp_service/app/viewsets.py
```diff
@@ -138,3 +142,24 @@ def get_queryset(self):
...
+        repo_version_pk = serialized_data.data["repo_version"]
+        task = dispatch(check_content, kwargs={"repo_version_pk": repo_version_pk})
+        return OperationPostponedResponse(task, request)
+
+
+class VulnerabilitiesDetail(APIView):
+    def get(self, request, uuid):
+        queryset = VulnerabilityReport.objects.get(id=uuid)
+        serializer = VulnerabilityReportSerializer(queryset, many=False)
+        return Response(serializer.data)
```
bmbouter, line 165: You totally nailed this part! Great job.

pulp_service/pulp_service/app/urls.py
```diff
@@ -1,14 +1,23 @@
...
 urlpatterns = [
     path("api/pulp-admin/", admin_site.urls),
     path("api/pulp/redirect-check/", RedirectCheck.as_view()),
     path("api/pulp/internal-server-error-check/", InternalServerErrorCheck.as_view()),
     path("api/pulp/raise-exception-check/", InternalServerErrorCheckWithException.as_view()),
     path("api/pulp/debug_auth_header/", DebugAuthenticationHeadersView.as_view()),
-    path("api/pulp/admin/tasks/", TaskViewSet.as_view({'get': 'list'})),
+    path("api/pulp/admin/tasks/", TaskViewSet.as_view({"get": "list"})),
+    path("api/pulp/vuln_report/", Vulnerabilities.as_view()),
+    path("api/pulp/vuln_report/<uuid:uuid>/", VulnerabilitiesDetail.as_view()),
```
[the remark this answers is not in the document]
bmbouter: This is an optional thing, but I'm really enjoying sharing the design stuff with you. You could have just `Vulnerabilities` (maybe renamed to `VulnerabilityReport` same as the model). It could have:

```
class VulnerabilityReport(APIView):

def list():   # the existing "Vulnerabilities.get()" view
  ...

def get():   # the existing "VulnerabilitiesDetail.get()" view
  ...

def post():   # the existing "Vulnerabilities.post()" view
  ...
```

Then for the urls.py you could have this:

```
    path("api/pulp/vuln_report/", VulnerabilityReport.as_view(TaskViewSet.as_view({"get": "list", "post": "post"})),
```

Why is this better? It's debatable if it's better, but the idea is to have fewer objects that do more things makes more powerful objects from an object oriented perspective I think that's a win.

pulp_service/pulp_service/app/serializers.py
```diff
@@ -13,9 +18,37 @@ class FeatureContentGuardSerializer(ContentGuardSerializer, GetOrCreateSerialize
...
+    A serializer for package scan.
+    """
+
+    repo_version = serializers.CharField()
+
+    def validate_repo_version(self, value):
+        try:
+            repo_version = RepoVersionHrefPrnFilter.get_repository_version(value)
+        except:
+            raise serializers.ValidationError(_("No matching model instance found."))
```
bmbouter, line 53:
```suggestion
            raise serializers.ValidationError(_("No matching RepositoryVersion instance found."))
```

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -0,0 +1,88 @@
...
+content_queue = Queue()
+
+
+async def check_content(repo_version_pk):
+    """
+    Get the list of contents from reop_version, build a package_list and makes an API request to
+    osv.dev using this package_list.
+    """
+    # create a thread to collect the Contents from RepositoryVersion
+    await asyncio.to_thread(_get_content_from_repo, repo_version_pk)
```
bmbouter, line 23: The use of `to_thread` needs to change for this new design. I should have pointed this out earlier, apologies.

The use of `await asyncio.to_thread` here will cause the thread to run until completion before this continues. This causes the background thread to not actually run concurrently so what that means is the `_get_content_from_repo` fills the entire queue and puts the `None` at the end before this coroutine, i.e. `check_content` can resume.

For concurrent operation you should just spawn a thread traditionally.

```
import threading
background_thread = threading.Thread(target=_get_content_from_repo, args=(repo_version_pk,))
background_thread.start()
```

There should be no reason to "join the thread" at the bottom because MainThread logically should continue to pull from the queue until `None` is received and the last thing `background_thread` does before exiting is emit a `None`. You could call `background_thread.join()` at the end of this coroutine just to be sure of a clean threading shutdown if you want though. Probably a good idea...

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -0,0 +1,88 @@
...
+    await _scan_packages()
+
+
+async def _scan_packages():
+    """
+    Makes a request to the osv.dev API and store the results in VulnerabilityReport model.
+    """
+    scanned_packages = {}
+    async with aiohttp.ClientSession() as session:
+        for package in _define_osv_package_list():
```
bmbouter, line 33: This could be simplified some if you instead of using `_define_osv_package_list()` as a helper function you use `iter` like this. Give this a try:

`for package in iter(content_queue.get, None):`

You can see some about how this works with in [this SO post](https://stackoverflow.com/questions/21157739/how-to-iterate-through-a-python-queue-queue-with-a-for-loop-instead-of-a-while-l).

I would then turn `_define_osv_package_list()` into a helper function which just helps build the `osv_data` object you use in the request. `_define_osv_package_list` already does just, this would just remove the queue fetching part. Also with it's changing role I'd probably rename to `_build_osv_data` or something. Maybe `_define_osv_package_list` is right. Totally up to you.

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -0,0 +1,88 @@
...
+            },
+            "content": content,
+        }
+
+
+def _identify_package_ecosystem(content: any) -> str:
+    """
+    Returns an osv.dev ecosystem (string) based on the content_type
+    """
+    model_type = content.TYPE
```
bmbouter, line 85: Consider removing this default value since we're going to want to explicitly add osv.dev support type-by-type later.

Instead I'd remove this line and add an `else` so that if none of the "supported content types" (in this case just NPMPackage) are not matched it raises a Python exception. Basically don't let unsupported types pass silently I think would be good.

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -0,0 +1,84 @@
...
+def _identify_package_ecosystem(content: any) -> str:
+    """
+    Returns an osv.dev ecosystem (string) based on the content_type
+    """
+    if isinstance(content, NPMPackage):
+        return getattr(PKG_ECOSYSTEM, "npm", None)
+    elif content.TYPE == "python":
+        return getattr(PKG_ECOSYSTEM, content.TYPE, None)
+    else:
+        raise RuntimeError("Package type not supported!")
```
bmbouter, line 84: great

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -0,0 +1,84 @@
...
+    background_thread.join()
+
+
+async def _scan_packages():
+    """
+    Makes a request to the osv.dev API and store the results in VulnerabilityReport model.
+    """
+    scanned_packages = {}
+    async with aiohttp.ClientSession() as session:
+        for content in iter(content_queue.get, None):
```
bmbouter, line 35: :100:

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -0,0 +1,84 @@
...
+            package = _build_osv_data(content)
```
bmbouter, line 36: :+1:

bmbouter (approved):
This looks exactly like what I was hoping for. You did a really great job, and also kept coming back for more improvements. I really enjoyed our collaboration. Please merge when you can and promote to staging (at least).

Also please add some docs to the App Interface Repo. Basically whatever scripts you used for npm would be useful there to populate some test content, and then run a scan report, and give an example of the response. We can make a separate JIRA if you want one, just let me know.

Thank you!

PR #440, opened 2025-03-13
Analyze npm packages-lock.json dependencies

git-hyagi:
ref: https://issues.redhat.com/browse/PULP-319


pulp_service/pulp_service/app/constants.py
```diff
@@ -6,3 +6,29 @@
 PKG_ECOSYSTEM = SimpleNamespace(
     rpm="Red Hat", npm="npm", maven="Maven", python="PyPI", gem="RubyGems"
 )
+
+# Define a basic schema for package-lock.json
+# https://docs.npmjs.com/cli/v11/configuring-npm/package-lock-json#file-format
```
bmbouter, line 11: :+1: way to reference the source (and have it be based on the source)

pulp_service/pulp_service/app/urls.py
```diff
@@ -19,4 +20,5 @@
     path("api/pulp/admin/tasks/", TaskViewSet.as_view({"get": "list"})),
     path("api/pulp/vuln_report/", VulnerabilityReport.as_view({"get": "list", "post": "post"})),
     path("api/pulp/vuln_report/<uuid:uuid>/", VulnerabilityReport.as_view({"get": "get"})),
+    path("api/pulp/npm_report/", NPMDependencyVulnerabilityReport.as_view({"post": "post"})),
```
bmbouter, line 23:
```suggestion
    path("api/pulp/npm_report/", NPMDependencyVulnerabilityReport.as_view()),
```

I can't remember, but you may be able to just get away with this alternative (and in the line above).

pulp_service/pulp_service/app/serializers.py
```diff
@@ -52,3 +57,27 @@ def validate_repo_version(self, value):
...
+    """
+    A serializer for npm package-lock.json files
+    """
+
+    package_json = serializers.FileField()
+
+    def verify_file(self):
+        uploaded_file = self.validated_data["package_json"]
+        temp_file = PulpTemporaryFile.init_and_validate(uploaded_file, None)
+        temp_file.save()
```
bmbouter, line 72: Is saving it necessary? I'm not sure of the types of data you get from `uploaded_file` versus `lock_file_content`. Maybe try validating on the data you get back directly from `self.validated_data["package_json"]`.

Also if the only use of saving is to read it back in (again I don't think you need to) then you don't need `PulpTemporaryFile` because the same process is writing/reading the file. `PulpTemporaryFile` is used for saving something to pass between processes, e.g. pulpcore-api and pulpcore-worker.

If you do need to save something and read it back in, you can use `tempfile.TemporaryFile` which creates an unnamed temp file. An unamed temp file isn't available on the filesystem, but is file-like in that it would provide `read()` like you need.

Best case though, don't save the file at all.

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -14,13 +14,20 @@
 content_queue = Queue()
 
 
-async def check_content(repo_version_pk):
+async def check_content(repo_version_pk=None, npm_package=None):
```
bmbouter, line 17: I know this is from the other PR, but I'd probably rename this to `check_repo_version` and leave this function to only check the repo version. Saving the few lines of code is not worth mixing the two calls together with an if/else. It increases the cyclomatic complexity and decreases the separation of concerns. You can go read about those topics if they are new. That's the motivation I offer in making a whole new function that does repeat some code.

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -59,10 +66,43 @@ def _get_content_from_repo(repo_version_pk: str):
...
+            )
+    content_queue.put(None)  # signal that there is no more content_units
+
+
 def _build_osv_data(content):
     """
     Helper function to build the osv.dev request data based on content object
     """
+
+    if isinstance(content, dict):
```
bmbouter, line 98: Similar here I'd not overload this with another if case to handle. Start by making it totally separate and then refactor the common parts into yet a new function that both functions call into (the reusable part).

pulp_service/pulp_service/app/urls.py
bmbouter: Also here I'd probably not introduce a new view here, but instead combine this view's functionality into `VulnerabilityReport.post`. That is a place where I'd increase it's concerns and make the view take either an npm_report, or a repo_version, not both, and not none. The end user is getting the same kind of report either way, is that right? If that's the case then we want them to use the same API endpoint and just make it take more options. Even if the reports differ somewhat it's ok, the JSON blob report we return is not a super-fixed schema in that it could have extra keys describing what the report is about, e.g. npm, vs python, vs a repo version.

pulp_service/pulp_service/app/viewsets.py
```diff
@@ -164,3 +165,14 @@ def post(self, request):
...
+
+
+class NPMDependencyVulnerabilityReport(ViewSet):
+
+    def post(self, request):
+        """Dispatch a task to scan the npm dependencies' vulnerabilities"""
+        serialized_data = NPMPackageLockSerializer(data=request.data)
+        serialized_data.is_valid(raise_exception=True)
+        file_content = serialized_data.verify_file()
+        task = dispatch(check_content, kwargs={"npm_package": file_content})
```
bmbouter, line 177: We try to avoid passing "large" data via arguments to tasks. This is actually where you want to use `PulpTemporaryFile` to save it and then load it in the tasking code.

pulp_service/pulp_service/app/serializers.py
```diff
@@ -52,3 +60,28 @@ def validate_repo_version(self, value):
...
+        try:
+            _logger.info(f"uploaded_file: {uploaded_file}")
+            _logger.info(f"temp_file: {temp_file.__dict__}")
+            lock_file_content_json = json.load(uploaded_file)
```
git-hyagi, line 79: **WIP**: the validation is failing here in ephemeral env, but it works in oci-env ...  I am investigating this ...

pulp_service/pulp_service/app/serializers.py
```diff
@@ -44,11 +49,43 @@ class ContentScanSerializer(serializers.Serializer):
...
+        required=False,
+        allow_null=True,
+        help_text=_("package-lock.json file with the definition of dependencies to be checked."),
+    )
+
+    def validate(self, data):
+        data = super().validate(data)
+        if bool(data.get("repo_version", None)) == bool(data.get("package_json", None)):
+            raise serializers.ValidationError(
+                _("Exactly one of 'repo_version' or 'package_json' must be specified.")
```
bmbouter, line 67: :heart:

pulp_service/pulp_service/app/urls.py
```diff
@@ -10,13 +11,18 @@
...
     path("api/pulp-admin/", admin_site.urls),
     path("api/pulp/redirect-check/", RedirectCheck.as_view()),
     path("api/pulp/internal-server-error-check/", InternalServerErrorCheck.as_view()),
     path("api/pulp/raise-exception-check/", InternalServerErrorCheckWithException.as_view()),
     path("api/pulp/debug_auth_header/", DebugAuthenticationHeadersView.as_view()),
     path("api/pulp/admin/tasks/", TaskViewSet.as_view({"get": "list"})),
-    path("api/pulp/vuln_report/", VulnerabilityReport.as_view({"get": "list", "post": "post"})),
-    path("api/pulp/vuln_report/<uuid:uuid>/", VulnerabilityReport.as_view({"get": "get"})),
+    path(
+        "api/pulp/vuln_report/", VulnerabilityReport.as_view({"get": "retrieve", "post": "create"})
```
bmbouter, line 25: :heart:

pulp_service/pulp_service/app/urls.py
```diff
@@ -10,13 +11,18 @@
...
+    ),
+    path("", include(router.urls)),
```
bmbouter, line 27: I realized something in doing this. We need these urls to be "domainized". These existing urls are for us as admins and we have admin permissions to operate on them, but end-users don't. Also we need the data, e.g. when listing `VulnerabilityReport` objects to be isolated per-domain. As-such, we need to make the model and viewset domain-based...

I'm asking in upstream #pulp-dev on matrix some for examples...

pulp_service/pulp_service/app/serializers.py
```diff
@@ -44,11 +49,43 @@ class ContentScanSerializer(serializers.Serializer):
...
     def validate_repo_version(self, value):
         try:
             repo_version = RepoVersionHrefPrnFilter.get_repository_version(value)
         except:
             raise serializers.ValidationError(_("No matching RepositoryVersion instance found."))
         return repo_version.pk
+
+    def verify_file(self):
+        uploaded_file = self.validated_data["package_json"]
+        temp_file = PulpTemporaryFile.init_and_validate(uploaded_file, None)
```
bmbouter, line 80: I see you're saving the file as part of validation, but then using that `PulpTemporaryFile` that is already saved to pass to the task via pk. That sounds efficient.

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -14,13 +14,31 @@
...
+    """
+    Spawn a thread to parse the package-json.lock file and make a request to osv.dev API
+    with each dependency found.
+    """
+    # create a thread to collect the Contents from package-lock.json file
+    background_thread = threading.Thread(target=_parse_npm_pkg_dependencies, args=(npm_package,))
+    await _start_thread_and_run_scan(background_thread)
+
+
+async def _start_thread_and_run_scan(background_thread):
```
bmbouter, line 37: nice refactor!

pulp_service/pulp_service/app/models.py
```diff
@@ -174,3 +175,4 @@ class VulnerabilityReport(models.Model):
 
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     vulns = models.JSONField()
+    pulp_domain = models.ForeignKey("core.Domain", default=get_domain_pk, on_delete=models.PROTECT)
```
gerrod3, line 178: Not sure if you want this to be `PROTECT` for `on_delete`. It would require you to have to delete all VulnerabilityReports before deleting a domain. Maybe this is the behavior you want, but I think having `CASCADE` would be simpler (there are already protections on other user objects)

pulp_service/pulp_service/app/viewsets.py
```diff
@@ -148,19 +151,33 @@ def view_name(cls):
...
+    def retrieve(self, request):
+        queryset = VulnReport.objects.filter(pulp_domain=get_domain())
         serializer = VulnerabilityReportSerializer(queryset, many=True)
         return Response(serializer.data)
 
-    def get(self, request, uuid):
-        queryset = VulnReport.objects.get(id=uuid)
+    def list(self, request, uuid):
+        try:
+            queryset = VulnReport.objects.get(id=uuid, pulp_domain=get_domain())
```
gerrod3, line 161: `id` is unique regardless of domain so you don't need to specify `pulp_domain` for this `get`.
```suggestion
            queryset = VulnReport.objects.get(id=uuid)
```

pulp_service/pulp_service/app/serializers.py
```diff
@@ -26,29 +32,62 @@ class Meta(ContentGuardSerializer.Meta):
...
 
-class VulnerabilityReportSerializer(ModelSerializer):
+class VulnerabilityReportSerializer(ModelSerializer, ValidateFieldsMixin):
     """
     A serializer for the VulnerabilityReport Model.
     """
 
     id = serializers.UUIDField()
     vulns = serializers.JSONField()
+    pulp_domain = serializers.CharField()
```
gerrod3, line 42: Typically we don't add pulp_domain to the serializer since the endpoints are domainized so we know what domain the object is in since we have specify it for the endpoint!

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -32,44 +51,69 @@ async def _scan_packages():
...
+def _parse_npm_pkg_dependencies(package_lock_content):
+    """
+    Parse the package-lock.json file to extract the packages[name][dependencies] and
+    add them to content_queue Queue
+
+    notes:
+    - we are striping the "~" and "^" from versions because osv.dev has no support to version range
+    - the old/legacy packages[dependencies] field is not supported
+    """
+    temp_file = PulpTemporaryFile.objects.get(pk=package_lock_content, pulp_domain=get_domain())
```
gerrod3, line 92: Same deal here with this `get`.
```suggestion
    temp_file = PulpTemporaryFile.objects.get(pk=package_lock_content)
```

pulp_service/pulp_service/app/urls.py
```diff
@@ -10,13 +11,23 @@
     VulnerabilityReport,
 )
 
+router = SimpleRouter(trailing_slash=False)
+router.register(
+    r"^api/pulp/(?P<pulp_domain>.+)/vuln_report/(?P<uuid>.+)/",
+    VulnerabilityReport,
+    basename="vuln-report",
+)
+
```
gerrod3, line 20: Is this duplicated with the `path` below?

pulp_service/pulp_service/app/viewsets.py
```diff
@@ -148,19 +151,33 @@ def view_name(cls):
...
-    def list(self, request):
-        queryset = VulnReport.objects.all()
+    def retrieve(self, request):
+        queryset = VulnReport.objects.filter(pulp_domain=get_domain())
         serializer = VulnerabilityReportSerializer(queryset, many=True)
         return Response(serializer.data)
 
-    def get(self, request, uuid):
-        queryset = VulnReport.objects.get(id=uuid)
+    def list(self, request, uuid):
```
gerrod3, line 159: I think you have these functions named backwards.

pulp_service/pulp_service/app/viewsets.py
```diff
@@ -148,19 +151,33 @@ def view_name(cls):
...
+    def create(self, request):
         serialized_data = ContentScanSerializer(data=request.data)
         serialized_data.is_valid(raise_exception=True)
-        repo_version_pk = serialized_data.data["repo_version"]
-        task = dispatch(check_content, kwargs={"repo_version_pk": repo_version_pk})
+
+        """Dispatch a task to scan the Content Units from a Repository"""
+        if repo_version_pk := serialized_data.data.get("repo_version", None):
+            dispatch_task, kwargs = check_content_from_repo_version, {
+                "repo_version_pk": repo_version_pk
```
gerrod3, line 174: If you specify a repo-version you should probably use a shared lock on the repository to prevent the task from running on a potentially deleted repo/repo-version.

gerrod3 (changes requested):

pulp_service/pulp_service/app/viewsets.py
```diff
@@ -148,19 +151,33 @@ def view_name(cls):
...
         serializer = VulnerabilityReportSerializer(queryset, many=True)
         return Response(serializer.data)
 
-    def get(self, request, uuid):
-        queryset = VulnReport.objects.get(id=uuid)
+    def list(self, request, uuid):
+        try:
+            queryset = VulnReport.objects.get(id=uuid, pulp_domain=get_domain())
+        except ObjectDoesNotExist:
+            raise Http404(f"No VulnerabilityReport found with uuid {uuid}")
```
bmbouter, line 163: please i8ln with `_()`

pulp_service/pulp_service/app/viewsets.py
bmbouter: +1 to this

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -32,44 +53,72 @@ async def _scan_packages():
...
                     )
                     scanned_packages[package_name] = json_body["vulns"]
-    await sync_to_async(VulnerabilityReport.objects.get_or_create)(
-        vulns=scanned_packages,
+    vuln_report, created = await sync_to_async(VulnerabilityReport.objects.get_or_create)(
+        vulns=scanned_packages, pulp_domain=get_domain()
     )
+    if created:
+        resource = await sync_to_async(CreatedResource)(content_object=vuln_report)
+        await sync_to_async(resource.save)()
```
gerrod3, line 72:
```suggestion
        await CreatedResource.objects.acreate(content_object=vuln_report)
```

gerrod3 (changes requested):

pulp_service/pulp_service/app/viewsets.py
```diff
@@ -146,21 +147,29 @@ def view_name(cls):
...
+    def create(self, request):
+        serialized_data = ContentScanSerializer(data=request.data, context={'request': request})
         serialized_data.is_valid(raise_exception=True)
-        repo_version_pk = serialized_data.data["repo_version"]
-        task = dispatch(check_content, kwargs={"repo_version_pk": repo_version_pk})
+
+        shared_resources = None
+        """Dispatch a task to scan the Content Units from a Repository"""
+        if repo_version_hyperlink := serialized_data.data.get("repo_version", None):
+            repository_version = repo_version_hyperlink.obj
```
gerrod3, line 163: After you call `is_valid` you typically get the validated data using `serializer.validated_data.get("repo_version")` and this returned value will be a RepositoryVersion (at least it should be)

pulp_service/pulp_service/app/serializers.py
```diff
@@ -31,24 +37,48 @@ class VulnerabilityReportSerializer(ModelSerializer):
...
+        uploaded_file = self.validated_data["package_json"]
+        temp_file = PulpTemporaryFile.init_and_validate(uploaded_file, None)
         try:
-            repo_version = RepoVersionHrefPrnFilter.get_repository_version(value)
-        except:
-            raise serializers.ValidationError(_("No matching RepositoryVersion instance found."))
-        return repo_version.pk
+            lock_file_content_json = json.load(uploaded_file)
+            validate(instance=lock_file_content_json, schema=NPM_PACKAGE_LOCK_SCHEMA)
+            temp_file.save()
```
gerrod3, line 78: Nice! Was there any need to `seek` the uploaded_file back to 0 before calling `json.load`? I notice it is no longer here and I know `init_and_validate` reads through the whole file first to calculate digests.

gerrod3 (approved):
Looks good!

pulp_service/pulp_service/app/migrations/0006_vulnerabilityreport.py
```diff
@@ -1,21 +1,32 @@
...
 
 from django.db import migrations, models
-import uuid
+import django.db.models.deletion
+import django_lifecycle.mixins
+import pulpcore.app.models.base
+import pulpcore.app.util
 
 
 class Migration(migrations.Migration):
```
bmbouter, line 10: Did this migration get promoted into production at any point? If so we can't modify it and a new migration 0007 would need to be made. If not, then this is ok. Please let us know.

pulp_service/pulp_service/app/serializers.py
```diff
@@ -31,24 +37,48 @@ class VulnerabilityReportSerializer(ModelSerializer):
...
     """
 
-    id = serializers.UUIDField()
     vulns = serializers.JSONField()
+    pulp_href = IdentityField(view_name="vuln_report-detail")
 
     class Meta:
         model = VulnerabilityReport
-        fields = ["id", "vulns"]
+        fields = ["vulns", "pulp_href"]
```
bmbouter, line 45: * Use a set instead of a list (because it's more efficient and doesn't need modification which a list would provide.
* Have the fields also included from the parent like [this line does](https://github.com/pulp/pulpcore/blob/7e9a1fdd294b014005b8dffea9579b126b130bb6/pulpcore/app/serializers/domain.py#L490). If you do that you don't need to define `pulp_href` at all because the parent does. Also you'll get `prn` and other useful fields in your serializer also.

pulp_service/pulp_service/app/serializers.py
bmbouter: +1 to this suggested change. Validate first, then save as `PulpTemporaryFile`

bmbouter (approved):
I requested some changes, but the real one to be sure of is the migration editing one. The others you can do or not do, or do as followup work. Please be sure of the migration is not in staging or production already before merging this

bmbouter:
@git-hyagi FYI if it turns out migration 6 was already run in either staging or production and you need to make a new migration, plan on making it migration 8 since I think @YasenT 's PR https://github.com/pulp/pulp-service/pull/452 will merge soon.

pulp_service/pulp_service/app/migrations/0007_alter_vulnerabilityreport_options_and_more.py
```diff
@@ -0,0 +1,47 @@
...
+import django.db.models.deletion
+import django.utils.timezone
+import pulpcore.app.util
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0130_upstreampulp_policy'),
+        ('service', '0006_vulnerabilityreport'),
```
bmbouter, line 14: migration 0007 is right. Thank you!

pulp_service/pulp_service/app/serializers.py
```diff
@@ -31,24 +37,48 @@ class VulnerabilityReportSerializer(ModelSerializer):
...
     """
 
-    id = serializers.UUIDField()
     vulns = serializers.JSONField()
+    pulp_href = IdentityField(view_name="vuln_report-detail")
 
     class Meta:
         model = VulnerabilityReport
-        fields = ["id", "vulns"]
+        fields = ModelSerializer.Meta.fields + ("vulns",)
```
bmbouter, line 45: :+1:

bmbouter:
These konflux errors look unrelated. This PR looks exactly what I think it needs to be. I'm merging.

PR #452, opened 2025-03-17
expose a new api endpoint for the ansible-logs-error script

YasenT:
New API endpoint to expose the ansible logs script, returning errors.
https://issues.redhat.com/browse/PULP-382

### DIFF
--- a/pulp_service/pulp_service/app/migrations/0008_ansiblelogreport.py
+++ b/pulp_service/pulp_service/app/migrations/0008_ansiblelogreport.py
@@ -11,7 +11,7 @@ class Migration(migrations.Migration):
 
     dependencies = [
         ('core', '0130_upstreampulp_policy'),
-        ('service', '0006_vulnerabilityreport'),
+        ('service', '0007_alter_vulnerabilityreport_options_and_more.py'),
     ]
 
     operations = [


pulp_service/pulp_service/app/models.py
```diff
@@ -174,3 +174,17 @@ class VulnerabilityReport(models.Model):
 
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     vulns = models.JSONField()
+    
+class AnsibleLogReport(models.Model):
```
decko, line 178: @YasenT We have a BaseModel that can be inherited here and avoid some duplication. Check it https://github.com/pulp/pulpcore/blob/18adef4f5c89d2c6c746fa635b50bb324259ee20/pulpcore/app/models/base.py#L20

pulp_service/pulp_service/app/urls.py
```diff
@@ -19,4 +20,6 @@
     path("api/pulp/admin/tasks/", TaskViewSet.as_view({"get": "list"})),
     path("api/pulp/vuln_report/", VulnerabilityReport.as_view({"get": "list", "post": "post"})),
     path("api/pulp/vuln_report/<uuid:uuid>/", VulnerabilityReport.as_view({"get": "get"})),
+    path("api/pulp/ansible-logs/", AnsibleLogAnalysisViewSet.as_view({"post": "create", "get": "list"})),
+    path("api/pulp/ansible-logs/<uuid:id>/", AnsibleLogAnalysisViewSet.as_view({"get": "retrieve"})),
```
bmbouter, line 24: I think they match top-to-bottom so the more specific one should come first. I'm 90% sure that's needed, but not 100%

pulp_service/pulp_service/app/viewsets.py
```diff
@@ -164,3 +168,46 @@ def post(self, request):
...
+        Returns a task that can be used to retrieve results.
+        """
+        serializer = AnsibleLogAnalysisSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        
+        log_url = serializer.validated_data['url']
+        role_filter = serializer.validated_data.get('role', ["ALL"])
+        
+        try:
+            task = dispatch_ansible_log_analysis(log_url, role_filter)
```
bmbouter, line 191: stylistically can the dispatch actually occur here instead of in a sub-function. The Pulp viewsets almost always do the dispatch themselves.

pulp_service/pulp_service/app/serializers.py
```diff
@@ -52,3 +52,30 @@ def validate_repo_version(self, value):
...
+    """
+    A serializer for Ansible log analysis requests.
+    """
+    url = serializers.URLField(
+        help_text=_("URL to the Ansible log file to be analyzed")
+    )
+    role = serializers.ListField(
+        child=serializers.CharField(),
+        help_text=_("List of roles to filter by, or ['ALL'] for all roles"),
+        default=["ALL"]
```
bmbouter, line 65: Oh the default is here too. This is probably the right place to have it, so then remove the other places its set.

pulp_service/pulp_service/app/models.py
```diff
@@ -18,7 +18,7 @@
 from django.contrib.postgres.fields import ArrayField
 
 from pulpcore.plugin.models import Domain
-from pulpcore.app.models import AutoAddObjPermsMixin, HeaderContentGuard
+from pulpcore.app.models import AutoAddObjPermsMixin, HeaderContentGuard, BaseModel
```
bmbouter, line 21: You didn't create this problem, but can you try switching the imports from `pulpcore.app` to `pulpcore.plugin`. Over-time directly importing from pulpcore other than `pulpcore.plugin` isn't safe from breaking changes. The "plugin API" though is safe which is offered inside `pulpcore.plugin`. Let me know if you have quesitons, or if you find you don't have all the objects you need.

bmbouter: Actually this is all-over our codebase. Let's do this as a separate activity not in this PR.

pulp_service/pulp_service/app/serializers.py
```diff
@@ -52,3 +52,30 @@ def validate_repo_version(self, value):
...
+    )
+
+class AnsibleLogReportSerializer(ModelSerializer):
+    """
+    A serializer for the AnsibleLogReport model.
+    """
+    id = serializers.UUIDField(read_only=True)
+    pulp_created = serializers.DateTimeField(read_only=True)
```
bmbouter, line 73: These are not needed and should be removed. Our project is kind of in transition from a "url" as the primary identifier to a "Pulp Resource Name" aka a "prn". That is inherited to your serializer from [here](https://github.com/pulp/pulpcore/blob/main/pulpcore/app/serializers/base.py#L447) so you don't need to have `id` or `pulp_created` here because you inherit from `ModelSerializer`.

pulp_service/pulp_service/app/tasks/ansible_log_parser.py
```diff
@@ -0,0 +1,196 @@
...
+    """
+    errors = await get_errors_from_ansible_log(role_filter, log_url)
+
+    # Store results in database
+    report = await sync_to_async(AnsibleLogReport.objects.create)(
+        log_url=log_url,
+        errors=errors,
+        error_count=len(errors),
+        role_filter=role_filter
+    )
```
bmbouter, line 55: At this point you should add a `CreatedResource` which is a Tasking field / concept where we report to the user what a task made. Without this, the user only would have the URL and would have to requery the API to find the `AnsibleLogReport`.

pulp_service/pulp_service/app/urls.py
bmbouter: If it all works, I don't have a preference.

pulp_service/pulp_service/app/viewsets.py
bmbouter: It's not how it's usually done, but I'm ok with that. I see where you're coming from. :+1:

pulp_service/pulp_service/app/models.py
bmbouter: https://issues.redhat.com/browse/PULP-391

pulp_service/pulp_service/app/serializers.py
```diff
@@ -52,3 +52,29 @@ def validate_repo_version(self, value):
         except:
             raise serializers.ValidationError(_("No matching RepositoryVersion instance found."))
         return repo_version.pk
+class AnsibleLogAnalysisSerializer(serializers.Serializer):
+    """
+    A serializer for Ansible log analysis requests.
+    """
+    url = serializers.URLField(
+        help_text=_("URL to the Ansible log file to be analyzed")
```
bmbouter, line 60: :+1:

pulp_service/pulp_service/app/serializers.py
```diff
@@ -52,3 +52,29 @@ def validate_repo_version(self, value):
...
+    """
+    domain = serializers.StringRelatedField(read_only=True)
+    log_url = serializers.URLField(read_only=True)
+    errors = serializers.JSONField(read_only=True)
+    error_count = serializers.IntegerField(read_only=True)
+    role_filter = serializers.JSONField(read_only=True)
+
+    class Meta:
+        model = AnsibleLogReport
+        fields = ['pulp_id','domain', 'pulp_created', 'log_url', 'errors', 'error_count', 'role_filter']
```
bmbouter, line 80:
```suggestion
        fields = ModelSerializer.Meta.fields + ('domain', 'log_url', 'errors', 'error_count', 'role_filter']
```

It's better to defer the field definition for inherited fields to the parent class. For example [here](https://github.com/pulp/pulpcore/blob/4e7b3394932567a75026237423caccd223a70621/pulpcore/app/serializers/domain.py#L490)

pulp_service/pulp_service/app/tasks/ansible_log_parser.py
```diff
@@ -0,0 +1,229 @@
...
+        # Use pulp_id instead of id - this is the key fix!
+        domain = await sync_to_async(Domain.objects.get)(pulp_id=domain_id)
+        report_data["domain"] = domain
+    
+    # Store results in database
+    
+    report = await sync_to_async(AnsibleLogReport.objects.create)(**report_data)
+    
+    # Register the created resource
+    await sync_to_async(CreatedResource.objects.create)(content_object=report)
```
bmbouter, line 68: In running your code locally, were you able to view the created resources on the task after it completed? I ask because @git-hyagi was having some challenges with that aspect so I wanted to see if you were able to get this working. From a code-read/review perspective, this all looks great.

pulp_service/pulp_service/app/tasks/ansible_log_parser.py
```diff
@@ -0,0 +1,229 @@
...
+            async with session.get(log_url, timeout=timeout) as response:
+                response.raise_for_status()
+                text = await response.text()
+                return text.splitlines()
+    except aiohttp.ClientError as e:
+        _logger.error("Failed to fetch log file: %s", str(e))
+        raise ValueError(f"Error fetching log file: {str(e)}")
+
+
+async def analyze_ansible_log(log_url, role_filter, domain_id=None):
```
bmbouter, line 38: With the default being specified in the dispatch function I recommend not having it be optional here, but clarifying in the docstring it can be UUID or None.

pulp_service/pulp_service/app/viewsets.py
```diff
@@ -164,3 +178,73 @@ def post(self, request):
         repo_version_pk = serialized_data.data["repo_version"]
         task = dispatch(check_content, kwargs={"repo_version_pk": repo_version_pk})
         return OperationPostponedResponse(task, request)
+
+class AnsibleLogReport(NamedModelViewSet):
```
bmbouter, line 182: :+1:

pulp_service/pulp_service/app/viewsets.py
```diff
@@ -164,3 +178,73 @@ def post(self, request):
...
+        else:
+            queryset = AnsibleLogReport.objects.filter(domain__isnull=True).order_by('-pulp_created')
+            
+        serializer = AnsibleLogReportSerializer(queryset, many=True)
+        return Response(serializer.data)
+
+    def retrieve(self, request, id, domain_name=None):
+        """Get a specific report within the domain."""
+        domain = self.get_domain()
+        set_domain(domain)
```
bmbouter, line 240: I don't think you need to call this in your viewsets. It's already done for you in an inbound middleware [here](https://github.com/pulp/pulpcore/blob/main/pulpcore/middleware.py#L49). You can then access the attribute on the `request` object directly instead.

bmbouter (changes requested):
I left some small changes. This is very close to do (I think). You're doing a great job. :+1:

pulp_service/pulp_service/app/viewsets.py
bmbouter: The middleware won't set the context for the task because middleware only runs on in bound HTTP requests. That's why passing it directly to your task is necessary. Still though, I think look into the issues some more because if you have to call it again, something isn't working as expected.

Did you try calling `get_domain()` or `get_domain_pk()` directly [from here](https://github.com/pulp/pulpcore/blob/main/pulpcore/app/util.py#L593-L597). If that still didn't work could you try debugging the pulpcore middleware call to verify it's running in your dev environment as expected?

pulp_service/pulp_service/app/models.py
```diff
@@ -174,3 +175,16 @@ class VulnerabilityReport(models.Model):
 
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     vulns = models.JSONField()
+    
+class AnsibleLogReport(BaseModel):
+    """
+    Model used to store Ansible log analysis results.
+    """
+    domain = models.ForeignKey("core.Domain", default=get_domain_pk, on_delete=models.CASCADE)
```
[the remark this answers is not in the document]
gerrod3:
```suggestion
    pulp_domain = models.ForeignKey("core.Domain", default=get_domain_pk, on_delete=models.CASCADE)
```
We try to keep the naming consistent across models.

pulp_service/pulp_service/app/serializers.py
```diff
@@ -52,3 +52,29 @@ def validate_repo_version(self, value):
...
+    """
+    pulp_domain = serializers.StringRelatedField(read_only=True)
+    log_url = serializers.URLField(read_only=True)
+    errors = serializers.JSONField(read_only=True)
+    error_count = serializers.IntegerField(read_only=True)
+    role_filter = serializers.JSONField(read_only=True)
+
+    class Meta:
+        model = AnsibleLogReport
+        fields = ['pulp_id', 'pulp_domain', 'pulp_created', 'log_url', 'errors', 'error_count', 'role_filter']
```
bmbouter, line 80: This would be slightly more efficient if it was a set (it's never modified). Also add to the set with the fields the parent serializer defines. So this would be `fields = ModelSerializer.fields + ('pulp_domain', ... )` similar to how [this other serializer does it](https://github.com/pulp/pulpcore/blob/7e9a1fdd294b014005b8dffea9579b126b130bb6/pulpcore/app/serializers/domain.py#L490) (as an example. That way (a) you can define less and (b) you get more useful fields that are very common across all pulp serializers, e.g. pulp_href, prn, pulp_created, etc.

pulp_service/pulp_service/app/viewsets.py
```diff
@@ -164,3 +181,68 @@ def post(self, request):
...
+        
+        Expects JSON with 'url' field.
+        Returns a task that can be used to retrieve results.
+        """
+        serializer = AnsibleLogAnalysisSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        
+        log_url = serializer.validated_data['url']
+        role_filter = serializer.validated_data.get('role')
+        domain = get_domain()
```
bmbouter, line 205: this seems unused

bmbouter (approved):
I left basically some nitpick comments. Fix that if you don't mind and then please merge. Great job on your first end-to-end application PR!!! :rocket:

pulp_service/pulp_service/app/serializers.py
```diff
@@ -52,3 +52,29 @@ def validate_repo_version(self, value):
...
+        child=serializers.CharField(),
+        help_text=_("List of roles to filter by, or ['ALL'] for all roles"),
+        default=["ALL"]
+    )
+
+class AnsibleLogReportSerializer(ModelSerializer):
+    """
+    A serializer for the AnsibleLogReport model.
+    """
+    pulp_domain = serializers.StringRelatedField(read_only=True)
```
gerrod3, line 72:
```suggestion
```
If the endpoint is scoped by domain then you know every returned result belongs to the domain you requested it from.

pulp_service/pulp_service/app/tasks/ansible_log_parser.py
```diff
@@ -0,0 +1,216 @@
...
+    Returns:
+        Task: The dispatched task object
+    """
+    # Create resources list for domain isolation
+    exclusive_resources = []
+    
+    try:
+        domain = get_domain()
+        # Add as a exclusive_resources
+        exclusive_resources.append(domain)
```
gerrod3, line 208: Does this need to be an exclusive resource? Every task that is dispatched has the domain as a shared resource by default. So if you specify that the domain should be an exclusive resource you are saying that it should block every other task in that domain.

pulp_service/pulp_service/app/viewsets.py
```diff
@@ -1,31 +1,48 @@
...
 from rest_framework.views import APIView
 from rest_framework.viewsets import ViewSet
+from rest_framework.mixins import ListModelMixin, RetrieveModelMixin
+from pulpcore.app.util import get_domain
+
 
+from pulpcore.app.models import Domain
 from pulpcore.app.response import OperationPostponedResponse
-from pulpcore.app.viewsets import ContentGuardViewSet, RolesMixin, TaskViewSet
+from pulpcore.app.util import set_domain
```
gerrod3, line 23:
```suggestion
```

pulp_service/pulp_service/app/viewsets.py
```diff
@@ -164,3 +181,68 @@ def post(self, request):
...
+            task = dispatch_ansible_log_analysis(log_url, role_filter)
+            return OperationPostponedResponse(task, request)
+        except Exception as e:
+            _logger.exception("Failed to dispatch log analysis task")
+            return Response(
+                {"error": _("Failed to analyze log: %(error)s") % {"error": str(e)}},
+                status=status.HTTP_500_INTERNAL_SERVER_ERROR
+            )
+    
+    def list(self, request):
```
gerrod3, line 217: Delete this method `list` and the method `retrieve`. Both are inherited from the `ListModalMixin` and `RetrieveModalMixn`.

gerrod3 (changes requested):
Didn't look too much into the task code, just that domains are used correctly. Also, I think there are a lot of unused imports still hanging around.

pulp_service/pulp_service/app/serializers.py
bmbouter: Tell me more about that please. We do this all over the pulp codebase so if there's an issue I think it means there's an underlying problem.

pulp_service/pulp_service/app/serializers.py
```diff
@@ -52,3 +52,28 @@ def validate_repo_version(self, value):
...
+    """
+    A serializer for Ansible log analysis requests.
+    """
+    url = serializers.URLField(
+        help_text=_("URL to the Ansible log file to be analyzed")
+    )
+    role = serializers.ListField(
+        child=serializers.CharField(),
+        help_text=_("List of roles to filter by, or ['ALL'] for all roles"),
+        default=["ALL"]
```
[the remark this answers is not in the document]
bmbouter: This actually came from the code we're porting from so I can see why @YasenT used this. Let's confirm with the stakeholder with their desire and go with that.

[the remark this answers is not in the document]
richm: Sure - use `*` - that's fine - if you are using the check_logs.py code, there are a couple of places like `args.role != ["ALL"]` which will need to be changed

pulp_service/pulp_service/app/migrations/0007_ansiblelogreport.py
gerrod3, line 1: I think this will need to be renumbered because Humberto's PR merged first.

gerrod3 (changes requested):
Rebase and ensure the migrations have a proper order, then lgtm!

pulp_service/pulp_service/app/tasks/ansible_log_parser.py
```diff
@@ -0,0 +1,205 @@
...
+    }
+    
+    # Store results in database
+    report = await sync_to_async(AnsibleLogReport.objects.create)(**report_data)
+    
+    # Register the created resource
+    await sync_to_async(CreatedResource.objects.create)(content_object=report)
+
+
+async def get_errors_from_ansible_log(role_filter, log_url):
```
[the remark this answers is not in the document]
bmbouter: @richm yes it's the same code, you linked us to. Whenever you go to use an AI we expect to have that new code replace this one. Is that also your understanding?

PR #465, opened 2025-03-21
Add delete to vulnreport endpoint

git-hyagi:
ref: https://issues.redhat.com/browse/PULP-414

### DIFF
--- a/pulp_service/pulp_service/app/viewsets.py
+++ b/pulp_service/pulp_service/app/viewsets.py
@@ -13,7 +13,7 @@ from rest_framework import status
 from rest_framework.exceptions import APIException
 from rest_framework.response import Response
 from rest_framework.views import APIView
-from rest_framework.mixins import ListModelMixin, RetrieveModelMixin
+from rest_framework.mixins import DestroyModelMixin, ListModelMixin, RetrieveModelMixin
 
 from pulpcore.app.response import OperationPostponedResponse
 from pulpcore.app.viewsets import ContentGuardViewSet, NamedModelViewSet, RolesMixin, TaskViewSet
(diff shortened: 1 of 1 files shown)


bmbouter:
There could be other errors in the background thread that cause it to die and could cause MainThread to deadlock... Consider this AI response and example about how to make this pattern more durable (several ways).

```
Handling exceptions in a background thread and ensuring the main thread does not deadlock when waiting on a queue can be tricky. Here's a structured approach to handle this gracefully:
Key Considerations:

    Exception Propagation: Ensure exceptions from the background thread are propagated to the main thread.

    Queue Timeout: Use a timeout when waiting on the queue to avoid indefinite blocking.

    Thread State Monitoring: Check if the background thread is still alive before waiting on the queue.

    Graceful Shutdown: Ensure the background thread can signal the main thread to stop waiting.

Implementation Example
python
Copy

import threading
import queue
import time
import sys

def background_task(q):
    try:
        for i in range(5):
            print(f"Background thread working: {i}")
            time.sleep(1)
            if i == 3:
                raise RuntimeError("Simulated error in background thread")
            q.put(i)
    except Exception as e:
        # Put the exception in the queue to notify the main thread
        q.put(e)
    finally:
        # Signal that the thread is done
        q.put(None)

def main():
    q = queue.Queue()
    thread = threading.Thread(target=background_task, args=(q,))
    thread.start()

    while True:
        try:
            # Wait for items with a timeout to avoid deadlock
            item = q.get(timeout=1)
            if item is None:
                print("Background thread finished.")
                break
            elif isinstance(item, Exception):
                print(f"Exception in background thread: {item}")
                break
            else:
                print(f"Main thread received: {item}")
        except queue.Empty:
            # Check if the background thread is still alive
            if not thread.is_alive():
                print("Background thread died unexpectedly.")
                break

    thread.join()  # Ensure the thread is cleaned up
    print("Main thread exiting.")

if __name__ == "__main__":
    main()

Explanation:

    Background Thread:

        The background thread performs its tasks and puts results into the queue.

        If an exception occurs, it puts the exception into the queue.

        It signals completion by putting None into the queue.

    Main Thread:

        Uses q.get(timeout=1) to wait for items with a timeout, avoiding indefinite blocking.

        Checks if the background thread is still alive if the queue is empty.

        Handles exceptions from the background thread by breaking the loop.

        Ensures the background thread is joined before exiting.

Key Points:

    Timeout: Using a timeout in q.get() ensures the main thread doesn't block indefinitely.

    Exception Handling: Exceptions from the background thread are propagated to the main thread via the queue.

    Graceful Shutdown: The background thread signals completion by putting None into the queue.

This approach ensures the main thread can handle exceptions and avoid deadlocks gracefully.
```

bmbouter:
@git-hyagi if you want to move the bugfix to its own PR to get the delete in more quickly feel free. That part looks fine. Or not, it's up to you.

git-hyagi:
@bmbouter thank you for all the explanations and help!!
I am going to move the bugfix to its own PR, since it is not that simple.

bmbouter (approved):
Thank you!

Issue #464, opened 2025-03-21
Task stuck in running state if file not found


git-hyagi:
**Describe the bug**
Trying to run the vulnerability report in stage env with **File Storage** will create a task that will raise an exception and will be stuck in "_running"_ state.

**To Reproduce**
Make a request to the vuln_report passing a lock file and using a domain that is **not** using object storage:
```
/api/pulp/${PULP_DOMAIN}/api/v3/vuln_report/ -F "package_json=@package-lock.json"
```

pulpcore-worker logs:
```
pulp [9849213c44a44f6f836bd01d858d54cc]: pulpcore.tasking.tasks:INFO: Starting task 0195b946-4a8b-77fe-80bd-aff95772af77 in domain: hyagi
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/storages/backends/s3.py", line 532, in _open
    f = S3File(name, mode, self)
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/storages/backends/s3.py", line 132, in __init__
    self.obj.load(**params)
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/boto3/resources/factory.py", line 565, in do_action
    response = action(self, *args, **kwargs)
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/boto3/resources/action.py", line 88, in __call__
    response = getattr(parent.meta.client, operation_name)(*args, **params)
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/botocore/client.py", line 570, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/botocore/context.py", line 124, in wrapper
    return func(*args, **kwargs)
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/botocore/client.py", line 1031, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (404) when calling the HeadObject operation: Not Found

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib64/python3.9/threading.py", line 980, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/sentry_sdk/integrations/threading.py", line 101, in run
    return _run_old_run_func()
[...]

PR #469, opened 2025-03-24
Fix task stuck in running state if file not found

git-hyagi:
fixes: #464

### DIFF
--- a/pulp_service/pulp_service/app/constants.py
+++ b/pulp_service/pulp_service/app/constants.py
@@ -32,3 +32,6 @@ NPM_PACKAGE_LOCK_SCHEMA = {
     },
     "required": ["name", "lockfileVersion", "packages"],
 }
+
+# timeout when waiting on tasks scan thread queue to avoid indefinite blocking.
+VULNERABILITY_TASK_THREAD_TIMEOUT = 60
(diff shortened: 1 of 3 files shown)


pulp_service/pulp_service/app/constants.py
```diff
@@ -32,3 +32,6 @@
     },
     "required": ["name", "lockfileVersion", "packages"],
 }
+
+# timeout when waiting on tasks scan thread queue to avoid indefinite blocking.
+VULNERABILITY_TASK_THREAD_TIMEOUT = 60
```
bmbouter, line 37: :+1:

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -78,7 +93,10 @@ def _get_content_from_repo_version(repo_version_pk: str):
     repo_version = RepositoryVersion.objects.get(pk=repo_version_pk)
     for content_unit in repo_version.content:
         content = content_unit.cast()
-        ecosystem = _identify_package_ecosystem(content)
+        try:
+            ecosystem = _identify_package_ecosystem(content)
+        except Exception as e:
+            content_queue.put(e)
```
bmbouter, line 99: Consider making a decorator that catches any exception, puts it into the content_queue.put() and the re-raises it and then decorate both `_get_content_from_repo_version` and `_parse_npm_pkg_dependencies`. I think it would be better to have the "exception propogation to MainThread" cover the entire background thread code, not just the errors you expect. Decorating "the whole thread's entrypoint" would do that in a clean, reusable way.

bmbouter (approved):
Thank you!

PR #470, opened 2025-03-25
Add support to gem in vuln report endpoint

git-hyagi:
ref: https://issues.redhat.com/browse/PULP-357

### DIFF
--- a/pulp_service/pulp_service/app/tasks/package_scan.py
+++ b/pulp_service/pulp_service/app/tasks/package_scan.py
@@ -144,7 +144,7 @@ def _identify_package_ecosystem(content: any) -> str:
     """
     if isinstance(content, NPMPackage):
         return getattr(PKG_ECOSYSTEM, "npm", None)
-    elif content.TYPE == "python":
+    elif content.TYPE in ["python", "gem"]:
         return getattr(PKG_ECOSYSTEM, content.TYPE, None)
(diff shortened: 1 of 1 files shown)


pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -144,7 +144,7 @@ def _identify_package_ecosystem(content: any) -> str:
     """
     if isinstance(content, NPMPackage):
         return getattr(PKG_ECOSYSTEM, "npm", None)
-    elif content.TYPE == "python":
+    elif content.TYPE in ["python", "gem"]:
```
bmbouter, line 147: nice! was this all that was needed? lol

git-hyagi: yep... we kind of let some types already prepared:
https://github.com/pulp/pulp-service/blob/f612a7d6a8a47ca7b6a1b38d0918156379b14b42/pulp_service/pulp_service/app/constants.py#L6-L8
so, since the endpoint and request format is the same for `osv.dev` and also for the `vuln_report` endpoint in our end, we didn't need to do more changes

bmbouter: This is nice. The cost of this additional feature is very low.

PR #471, opened 2025-03-25
Test if the response is a simple Response object.

decko:


git-hyagi:
After an internal investigation, I think we are fine to close this PR in favor of: https://github.com/pulp/pulp-service/pull/473

PR #472, opened 2025-03-25
Handle osv.dev request report pagination

git-hyagi:
https://google.github.io/osv.dev/post-v1-query/#pagination
> _The OSV.dev API uses pagination for queries that return a large number of vulnerabilities. When pagination is used, the next_page_token is given in the response, indicating that there are more results to return. You will need to run additional queries using the page_token to see the remaining results, repeating queries until the next_page_token is no longer included in the response._

### DIFF
--- a/pulp_service/pulp_service/app/tasks/package_scan.py
+++ b/pulp_service/pulp_service/app/tasks/package_scan.py
@@ -68,12 +68,22 @@ async def _scan_packages(background_thread):
                 async with session.post(url=OSV_QUERY_URL, data=data) as response:
                     response_body = await response.text()
                     json_body = json.loads(response_body)
+                    osv_package_name = osv_data["package"]["name"]
+                    osv_package_version = osv_data["version"]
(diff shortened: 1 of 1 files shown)


pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -68,12 +68,22 @@ async def _scan_packages(background_thread):
...
-                        )
                         scanned_packages[package_name] = json_body["vulns"]
+                    if next_page_token := json_body.get("next_page_token"):
+                        next_page_request = _build_osv_data(
+                            osv_package_name,
+                            osv_data["package"]["ecosystem"],
+                            osv_package_version,
+                            next_page_token,
+                        )
+                        content_queue.put(next_page_request)
```
bmbouter, line 86: dude, way to use the `content_queue` design!

bmbouter (approved):
Thank you!

PR #481, opened 2025-04-03
Vuln report rpm

git-hyagi:


git-hyagi:
investigating the ci failed ... I guess we will need to backport https://github.com/pulp/pulp_rpm/pull/3928 and release it.

pulp_service/pulp_service/app/constants.py
```diff
@@ -7,6 +7,10 @@
     rpm="Red Hat", npm="npm", maven="Maven", python="PyPI", gem="RubyGems"
 )
 
+# Labels used in RPM repositories for vulnerability report tasks
+OSV_RH_ECOSYSTEM_LABEL = "osv_dev ecosystem"
+OSV_RH_ECOSYSTEM_CPES_LABEL = "osv_dev cpes"
```
bmbouter, line 12:
```suggestion
OSV_RH_ECOSYSTEM_CPES_LABEL = "osv.dev cpes"
```

Is using a period an option?

git-hyagi: Unfortunately, not: https://github.com/pulp/pulpcore/blob/9b6c47ffa88377de814c47c2ceb0cf819cf6f288/pulpcore/app/serializers/fields.py#L430

I also had some issues with pulp-cli: https://github.com/pulp/pulp-cli/issues/1165

pulp_service/pulp_service/app/serializers.py
```diff
@@ -63,12 +67,40 @@ class ContentScanSerializer(serializers.Serializer, ValidateFieldsMixin):
...
-        if bool(data.get("repo_version", None)) == bool(data.get("package_json", None)):
+        repo_ver = data.get("repo_version", None)
+        pkg_json = data.get("package_json", None)
+        if bool(repo_ver) == bool(pkg_json):
             raise serializers.ValidationError(
                 _("Exactly one of 'repo_version' or 'package_json' must be specified.")
             )
+
+        # no more validations needed for pkg_json
+        if pkg_json:
```
bmbouter, line 78: Consider using the walrus `:-` here

pulp_service/pulp_service/app/serializers.py
```diff
@@ -63,12 +67,40 @@ class ContentScanSerializer(serializers.Serializer, ValidateFieldsMixin):
...
+    def _validate_rpm_repo_expected_fields(self, repo):
+        """
+        verify if the label 'osv.dev ecosystem' == 'Red Hat'
+        verify if len(label['osv.dev cpe']) > 0
+        """
+        # for now, we are only supporting 'Red Hat' for rpms
+        if repo.pulp_labels.get(OSV_RH_ECOSYSTEM_LABEL) == "Red Hat":
+            # 'osv.dev cpe' must be provided for Red Hat ecosystem
+            return (
+                repo.pulp_labels.get(OSV_RH_ECOSYSTEM_CPES_LABEL)
```
bmbouter, line 99: You could maybe make this simpler+faster with the walrus also. It's a stylistic choice. Up to you.

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -150,13 +154,32 @@ def _build_osv_data(name, ecosystem, version=None, next_page_token=None):
...
+            return []
     else:
         raise RuntimeError("Package type not supported!")
+
+
+def _convert_rhel_repo_cpe(repo):
+    """
+    Convert the CPE into osv.dev expected format
+    """
+    cpe_prefix = r"^cpe:\/[oa]:redhat"
```
bmbouter, line 179: Since this function is called N times, it would be more efficient to move this outside of the function and make it `pattern = re.compile(r"your_regex_here")` then use it over and over.

pulp_service/pulp_service/tests/functional/test_vulnerability_report.py
```diff
@@ -146,3 +149,37 @@ def test_repo_version_vuln_report(
...
+            file_url=RPM_SAMPLE_PACKAGE_URL,
+            repository=repository.pulp_href,
+        ).task
+    )
+
+    latest_repo_version = rpm_repository_api.read(
+        rpm_rpm_repository_href=repository.pulp_href
+    ).latest_version_href
+    report = run_vuln_report(repo_version=latest_repo_version)
+    vuln_report_checks(report)
```
bmbouter, line 185: I think in addition to the general checks, you should assert on specific vulns reported. Since the test should run over and over I think you can safely assert that "at least" the same, known vulns are reported, even if additional vulns get reported later.

Please do the same with each content type vuln report test. Basically move the assertions on the vulns out of `vuln_report_checks` and into each content type test and have them be specific to that vuln report content type.

My apologies for missing this on the earlier tests PR and bringing it up here.

pulp_service/requirements.txt
```diff
@@ -1,5 +1,5 @@
 pulpcore==3.74.1
-pulp-rpm==3.28
+pulp-rpm==3.29.1
```
bmbouter, line 2: Maybe don't also do this in this PR.

bmbouter (changes requested):
I left a few comments, but the main request is for some assertion improvements in the tests.

pulp_service/requirements.txt
git-hyagi: fixed in https://github.com/pulp/pulp-service/pull/482

bmbouter (approved):
This looks great to me! Please promote to production and then record your demo from that environment. Thank you!

PR #483, opened 2025-04-04
Adds DomainBasedPermission class to mvn deploy API.

dkliban:


bmbouter:
Why does the upstream code even have those lines? I'm just curious.

If you say this fixes it Im +1 to merge.

dkliban:
The upstream code has that code because this viewset doesn't have a AccessPolicy defined.

bmbouter (approved):
Thank you!

PR #492, opened 2025-04-15
Patch the task logger to include the task name.

decko:


bmbouter (approved):
This looks good to me. Ty!

PR #493, opened 2025-04-16
Adds storage settings to OCI Storage Backend.

dkliban:
The storage backend now takes a username, password, and repository name.

It also now saves artifacts with their checksum as the tag in quay.

### DIFF
--- a/images/assets/patches/0011-ocistorage-backend-changes.patch
+++ b/images/assets/patches/0011-ocistorage-backend-changes.patch
@@ -1,32 +1,17 @@
-From 63ffa36a6eaf2c7cae84c91070228b6b1afd9bcf Mon Sep 17 00:00:00 2001
+From 216357d83b3061a2fbbea04a4110199e5a75e98b Mon Sep 17 00:00:00 2001
 From: Dennis Kliban <dkliban@redhat.com>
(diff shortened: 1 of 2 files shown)


images/assets/patches/0011-ocistorage-backend-changes.patch
```diff
@@ -1,32 +1,17 @@
...
-diff --git a/pulpcore/app/models/domain.py b/pulpcore/app/models/domain.py
-index 18dd3e892..2fc79e09f 100644
---- a/pulpcore/app/models/domain.py
-+++ b/pulpcore/app/models/domain.py
-@@ -39,7 +39,7 @@ class Domain(BaseModel, AutoAddObjPermsMixin):
-     description = models.TextField(null=True)
-     # Storage class is required, optional settings are validated by serializer
-     storage_class = models.TextField(null=False)
--    storage_settings = EncryptedJSONField(default=dict)
-+    storage_settings = EncryptedJSONField(default=dict, blank=True)
```
decko, line 24: Does it need to have a `blank=True` here?

dkliban: I removed the blank=True because the storage backend now requires settings so we don't need to support not providing settings.

PR #495, opened 2025-04-22
Handle errors pulling from quay.io in OCI Storage backend

dkliban:


bmbouter (approved):
Please close the bug after verifying in production. Thank you!

PR #502, opened 2025-04-25
Add a patch to block the task queue from receiving new tasks.

decko:


bmbouter (approved):
This looks great. Also do the resource changes happening here or is that in the MR on gitlab? If it's on gitlab then I'd say this is ready!

PR #506, opened 2025-05-02
Add some keywords to log entries to parse it easily

decko:
- **Set some keywords to easily parse logs when it's needed.**
- **Change the query to use the new parseable artifact_size info.**


bmbouter (approved):
Looks good to me.

PR #508, opened 2025-05-05
Add a new graph for cache hits.

decko:
image

Also, the proportion of the columns is corrected given more data.


bmbouter:
Is this a new panel or a new graph inside the existing panel? (I'm pretty unclear on the grafana terminology). Also if you can post a picture of what this produces on the PR that would be nice (but I'm ok if not). Thank you!

bmbouter (approved):
Thanks! Please promote to production when you can.

PR #522, opened 2025-05-23
Task Ingestion test plan

decko:
## Summary by Sourcery

Documentation:
- Add Task Ingestion Test Plan with sections covering current environment, key metrics, step-by-step test plan, results template, and example measurement script.

### DIFF
--- a/docs/test_plans/0002-Task_Ingestion.md
+++ b/docs/test_plans/0002-Task_Ingestion.md
@@ -1,12 +1,12 @@
 # Task Ingestion Test Plan
 
 <!--toc:start-->
-- [Worker Scalability Test Plan](#worker-scalability-test-plan)
+- [Task Ingestion Test Plan](#task-ingestion-test-plan)
   - [Current State](#current-state)
   - [Metrics to follow](#metrics-to-follow)
   - [Test Plan](#test-plan)
   - [Results](#results)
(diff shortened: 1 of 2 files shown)


docs/test_plans/0002-Task_Ingestion.md
```diff
@@ -0,0 +1,128 @@
...
+        int: Number of tasks processed, or None if timeout or error occurs.
+    """
+    with ThreadPoolExecutor(max_workers=1) as executor:
+        future = executor.submit(send_request_and_process_tasks, url)
+        try:
+            result = future.result()
+            return result
+        except TimeoutError:
+            print("Operation timed out after 25 seconds.")
+            return None
```
bmbouter, line 116: I think the client code should wait indefinitely and the time limits are entirely server side.

bmbouter: I think maybe I'm confused on what this is measuring, apologies if I am. What made sense to me was:
1. the client waits indefinitely
2. the threadpool executes a lot of "dispath and wait" functions (each in it's own thread) so the parallelization needs to be high on the threadpool
3. The timing element in this client code should be for the entire test duration so I expected tasks to be dispatched with a 25 second value, but the client code to have something like a 300 second (5 minute) value.

PR #523, opened 2025-05-23
Add a view to test the Task Ingestion.

decko:


pulp_service/pulp_service/app/tasks/util.py
```diff
@@ -37,4 +38,8 @@ def wrapper(*args, **kwargs):
                 # Re-raise the exception
                 raise
         return wrapper
-    return decorator
\ No newline at end of file
+    return decorator
+
+
+def create_pass_task():
```
bmbouter, line 44:
```suggestion
def no_op_task():
```

pulp_service/pulp_service/app/settings.py
```diff
@@ -8,4 +8,5 @@
 SUBSCRIPTION_API_URL = "https://subscription.stage.api.redhat.com/svcrest/subscription/v5/featureStatus"
 SUBSCRIPTION_API_CERT = ""
 AUTHENTICATION_HEADER_DEBUG = False
-INSTALLED_APPS = "@merge django.contrib.admin.apps.SimpleAdminConfig"
\ No newline at end of file
+INSTALLED_APPS = "@merge django.contrib.admin.apps.SimpleAdminConfig"
+TEST_TASK_INGESTION = False
```
bmbouter, line 12: This can be taken out (I think) because with it not being part of any domain, only an admin could authorize to run it with the middleware that's in place.

pulp_service/pulp_service/app/viewsets.py
```diff
@@ -181,3 +184,28 @@ def create(self, request):
...
+    permission_classes = []
+
+
+    def get(self, request=None, path=None, task_num=1):
+        if not settings.TEST_TASK_INGESTION:
+            raise PermissionError("Access denied.")
+
+        task_count = 0
+        start_time = datetime.now()
+        timeout = timedelta(seconds=25)
```
bmbouter, line 201: I believe 25 should come from the client. It's possible in different environments we'll set the gunicorn timeout differently. This should be a required paramenter passed to the viewset.

pulp_service/pulp_service/app/viewsets.py
```diff
@@ -181,3 +184,28 @@ def create(self, request):
...
         return OperationPostponedResponse(task, request)
+
+
+class TaskIngestionDispatcherView(APIView):
+
+    authentication_classes = []
+    permission_classes = []
+
+
+    def get(self, request=None, path=None, task_num=1):
```
bmbouter, line 195: task_num should have no default. If unset it should loop forever until task_timeout time.

pulp_service/pulp_service/app/viewsets.py
```diff
@@ -181,3 +184,28 @@ def create(self, request):
...
+
+        while datetime.now() < start_time + timeout:
+            dispatch(
+                'pulp_service.app.tasks.util.create_pass_task',
+                exclusive_resources=str(uuid4())
+            )
+                
+            task_count = task_count + 1
+
+        return Response({"tasks_executed": task_count})
```
bmbouter, line 211: :+1:

PR #524, opened 2025-05-23
List domains alphabetically on the DomainOrg form of django admin.

dkliban:
## Summary by Sourcery

Enhancements:
- Sort the DomainOrgForm domain dropdown by name for easier lookup


bmbouter (approved):
Overall I think this is right. I can't think of an easier way.

PR #530, opened 2025-06-02
Adds domains ManyToManyField to the DomainOrg model

dkliban:
This also adds a data migration that populates the domains field with existing domain field foreign keys.

This also updates the post_create_domain signal to populate the domains field with the foreign key of the domain being created.


bmbouter (approved):
Looks good to me. Please spot check the data copied on staging before promoting to production. ty!

PR #532, opened 2025-06-04
Draft for the Task Throughput Test Plan.

decko:


docs/test_plans/0003-Task_Throughput.md
```diff
@@ -0,0 +1,132 @@
...
+2. Using the logs, calculate the number of successful tasks, and the number of tasks completed per second.
+3. Calculate the Task Latency using a sample of executed tasks.
+
+
+## Results
+Date of the test: YYYY/mm/dd
+
+A possible template for the table could be:
+
+| Run  | Workers | Requests | Successful Tasks | Task Latency | Error Rates | Observations |
```
bmbouter, line 55:
```suggestion
| Run  | pulp-worker Count | Requests | Successful Tasks | Task Latency | Error Rates | Observations |
```

bmbouter: The `Tasks Submitted` is how you're loading the test. I think that number should be "large" like 10,000. Since we learned from the task ingestion test that we can do ~150 tasks / sec should take roughly 66 seconds.

Since the rate of incoming work is upper bound at ~150 tasks / sec, you only need to dispatch enough to determine if N pulpcore-workers can process work "that fast". I suspect they can, but this test will show us.

docs/test_plans/0003-Task_Throughput.md
```diff
@@ -0,0 +1,132 @@
...
+**Key:**
+- Use `-` for unavailable metrics.
+
+## An example to calculate the number of tasks ingested
+```python
+import requests
+import time
+from concurrent.futures import ThreadPoolExecutor, TimeoutError
+
+def send_request_and_process_tasks(url):
```
bmbouter, line 71: I think you want to basically use the code you used exactly from the ingestion test and pick a reasonable maximum of "ingestion API tasks" which from the previous data I read as N=2 as being a good choice. So in that scenario you'll have two pulpcore-api workers creating tasks like crazy, and the workers running in the background and picking them up. Just link to that code and call out how it's going to be run. You don't need to "rewrite" it here.

bmbouter: The new code here you need to:
1. Wait for all tasks to complete (you don't need to calculate as it runs). Or just manually check it and run your calculation code afterwards.
2. Find the start time and end time of the test. I recommend finding the very first task created_at time as the start time, and the very latest end_time as the end time. Use the difference in those timestamps as the duration of the test (time_in_seconds_duration).
3. Compute the rate from the number of tasks / time_in_seconds_duration.

bmbouter (approved):
This all looks good to me in terms of clarity and accuracy.

PR #540, opened 2025-06-10
Removes deprecated domain field from DomainOrg model.

dkliban:
## Summary by Sourcery

Remove the deprecated DomainOrg.domain field from the model and database, and update the admin interface to use the existing domains many-to-many field.

Enhancements:
- Drop the DomainOrg.domain field and remove its references in DomainOrgAdmin
- Adjust DomainOrgAdmin to display only user and org_id and replace the foreign key formfield override with a many-to-many override for domains

Chores:
- Add a Django migration to remove the domain field from the database schema


bmbouter (approved):
LGTM, ty!

PR #585, opened 2025-07-18
self service domain creation

YasenT:
The new endpoint would allow users to create domains with S3 config on demand, without our involvement.

## Summary by Sourcery

Add a self-service API endpoint to create S3-backed domains using a template domain's configuration

New Features:
- Introduce CreateDomainView at POST /api/pulp/create-domain/ to enable on-demand domain creation
- Automatically apply storage settings, storage class, and labels from the 'template-domain-s3' model domain

### DIFF
--- a/pulp_service/pulp_service/app/viewsets.py
+++ b/pulp_service/pulp_service/app/viewsets.py
@@ -17,7 +17,7 @@ from rest_framework import status
 from rest_framework.exceptions import APIException
 from rest_framework.response import Response
 from rest_framework.views import APIView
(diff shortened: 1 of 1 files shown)


bmbouter (commented):
This looks great! My only question is: what's the timeline for adding tests, do you want to add them in this PR or come back and add them later? My desire is a basic functional test, but I'm also not sure if environments like ephemeral (where tests are run) would have the `tempalte-domain-s3` data already or not. Details details.

I'm interested in what the rest of the services team would say about that. Also some link to existing tests that we carry (separate from upstream) would be helpful I think too.

I am ok with merging this if a ticket is created to add tests later and prioritized soon. Up to you all.

PR #626, opened 2025-08-14
Enable TASK_DIAGNOSTICS

dkliban:
This patch removes the custom middleware from pulp-service and enables the TASK_DIAGNOSTICS provided by pulpcore.


deploy/clowdapp.yaml
```diff
@@ -615,6 +615,8 @@ objects:
             value: ${PULP_ALLOWED_CONTENT_CHECKSUMS}
           - name: PULP_TASK_PROTECTION_TIME
             value: ${PULP_TASK_PROTECTION_TIME}
+          - name: PULP_TASK_DIAGNOSTICS
+            value: ${{PULP_TASK_DIAGNOSTICS}}
```
decko, line 619: Shouldn't this var be surrounded by single brackets instead of doubles?

decko (approved):

PR #640, opened 2025-08-26
Ignore the healthcheck endpoint and the status endpoint in the request duration metric.

decko:
## Summary by Sourcery

Enhancements:
- Introduce a new OTEL_PYTHON_EXCLUDED_URLS environment variable in clowdapp.yaml to filter out livez and status endpoints from OTEL metrics


bmbouter (approved):
I'm iffy on the benefit here versus the downside of not getting metrics data on these endpoints, but if it's a problem in practice we can adjust later. I'm lgtm for now. Ty!

PR #673, opened 2025-09-19
Add a task to set the group on all content sources DomainOrgs

decko:
## Summary by Sourcery

New Features:
- Add a Clowd task podSpec and job invocation to assign the 'contentsources' group to DomainOrg records labeled for content sources

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -980,6 +980,36 @@ objects:
               value: "/tmp"
             - name: XDG_CACHE_HOME
               value: "/tmp"
+      - name: set-contentsources-group-to-cs-domains
+        podSpec:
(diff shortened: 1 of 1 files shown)


deploy/clowdapp.yaml
```diff
@@ -980,6 +980,36 @@ objects:
...
+          image: ${IMAGE}:${IMAGE_TAG}
+          command: [ 'pulpcore-manager' ]
+          args:
+            - "shell"
+            - "-c"
+            - |
+              from os import getenv
+              from django.contrib.auth import get_user_model
+              cs_group = Group.objects.get(name="contentsources")
+              DomainOrg.objects.filter(domains__pulp_labels__contains={"contentsources": "true"}).update(group=cs_group)
```
bmbouter, line 994: I think we should also remove the user at the same time.

PR #692, opened 2025-09-29
Hide the User admin view when it's a normal user.

decko:
## Summary by Sourcery

Restrict access to the Django User admin for non-staff users by hiding the module and filtering permissions and querysets based on is_staff and is_superuser flags

Enhancements:
- Limit User admin queryset: superusers see all users; staff see only themselves and group members; others see none
- Permit change permissions only for staff users sharing groups with the target user
- Show the User admin module exclusively to staff and superuser accounts instead of all authenticated users

### DIFF
--- a/pulp_service/pulp_service/app/admin.py
+++ b/pulp_service/pulp_service/app/admin.py
@@ -97,13 +97,16 @@ class PulpUserAdmin(UserAdmin):
             return qs
 
         # Staff users can see themselves + users in their groups
-        user_groups = request.user.groups.all()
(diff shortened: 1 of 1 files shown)


bmbouter:
> I understood that we're only removing the limitation of the login for normal users. Since the code for the permissions was there, just branching seems a "quick wins" solution for the moment.

Yes, but the values of `is_staff` are uncertain. Instead I think re-arrange the if-statements to focus the check on `is_admin` which we can be sure of.

bmbouter (approved):
I think this is good, let's try this.

PR #1079, opened 2026-04-22
Add more Pulp Exceptions.

aKlimau:
Assisted-by: Claude Sonnet 4.6 <noreply@anthropic.com>

## Summary by Sourcery

Introduce dedicated Pulp exceptions for vulnerability scanning failures and unsupported package types.

Enhancements:
- Replace generic runtime errors in package scanning with specific PulpException-derived errors carrying structured messages and codes.
- Add a central exceptions module defining service-specific error types for vulnerability report threading issues and unsupported content types.

### DIFF
--- a/pulp_service/pulp_service/app/exceptions.py
+++ b/pulp_service/pulp_service/app/exceptions.py
@@ -1,0 +1,52 @@
+from gettext import gettext as _
+
+from pulpcore.plugin.exceptions import PulpException
+
+
+class VulnReportTaskError(PulpException):
+    """
+    Raised when the background vulnerability report thread fails.
+    """
+
+    error_code = "SVC0001"
+
+    def __init__(self, osv_data):
+        self.osv_data = osv_data
+
+    def __str__(self):
+        return f"[{self.error_code}] " + _(
+            "Background vuln report task failed to execute: {osv_data}"
+        ).format(osv_data=self.osv_data)
+
+
+class VulnReportThreadDiedError(PulpException):
+    """
+    Raised when the vulnerability report background thread dies unexpectedly.
+    """
+
+    error_code = "SVC0002"
+
+    def __str__(self):
+        return f"[{self.error_code}] " + _("Vuln report task thread died unexpectedly.")
+
+
+class VulnReportThreadTimeoutError(PulpException):
+    """
+    Raised when the vulnerability report background thread takes too long.
+    """
+
+    error_code = "SVC0003"
+
+    def __str__(self):
+        return f"[{self.error_code}] " + _("Background vuln report thread took too long.")
+
+
+class UnsupportedPackageTypeError(PulpException):
+    """
+    Raised when a content type cannot be mapped to a supported package ecosystem.
+    """
+
+    error_code = "SVC0004"
+
+    def __str__(self):
+        return f"[{self.error_code}] " + _("Package type not supported!")
--- a/pulp_service/pulp_service/app/tasks/package_scan.py
+++ b/pulp_service/pulp_service/app/tasks/package_scan.py
@@ -16,6 +16,12 @@ from pulp_service.app.constants import (
     PKG_ECOSYSTEM,
     VULNERABILITY_TASK_THREAD_TIMEOUT,
 )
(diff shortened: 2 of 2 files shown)


decko:
The `pulp-bonfire-tekton` integration test failure looks like an infrastructure flake — the `deploy-application` step timed out after ~15 minutes. The build itself (`pulp-on-pull-request`) passed fine. Could you re-trigger the pipeline? /retest

decko (commented):
Nice improvement — replacing bare `RuntimeError` with structured `PulpException` subclasses gives better error reporting through the Pulp task framework and enables i18n. The error code scheme (SVC0001–SVC0004) is clean and extensible.

A few things to address before merging — see inline comments for details:

1. **Bug**: `VulnReportTaskError.__init__` doesn't call `super().__init__()`, which may break `PulpException` initialization (error payload, DRF response handling).
2. **Bug**: Exception chaining is lost when raising `VulnReportTaskError` — the original traceback from the background thread is silently dropped.
3. **Suggestion**: `UnsupportedPackageTypeError` would be more useful if it included the actual content type that wasn't supported, for diagnostic purposes.

Also noting: the pipeline failure (`test_entitled_feature_service` returning 403) is unrelated to this PR — it's a feature service / ephemeral environment issue.

pulp_service/pulp_service/app/exceptions.py
```diff
@@ -0,0 +1,52 @@
...
+
+class VulnReportTaskError(PulpException):
+    """
+    Raised when the background vulnerability report thread fails.
+    """
+
+    error_code = "SVC0001"
+
+    def __init__(self, osv_data):
+        self.osv_data = osv_data
```
decko, line 14: This overrides `__init__` without calling `super().__init__()`. `PulpException` likely initializes fields that downstream handlers depend on (DRF error responses, task error recording, translation). The other three exceptions in this file inherit the base `__init__` and are fine, but this one breaks the chain.

```python
def __init__(self, osv_data):
    self.osv_data = osv_data
    super().__init__()
```

pulp_service/pulp_service/app/tasks/package_scan.py
```diff
@@ -69,7 +75,7 @@ async def _scan_packages(background_thread):
                 lambda: content_queue.get(timeout=VULNERABILITY_TASK_THREAD_TIMEOUT), None
             ):
                 if isinstance(osv_data, Exception):
-                    raise RuntimeError(f"Background vuln report task failed to execute: {osv_data}")
+                    raise VulnReportTaskError(osv_data=osv_data)
```
decko, line 78: The original exception context is lost here. `osv_data` at this point is an `Exception` instance (put into the queue by the `except_catch_and_raise` decorator), but without chaining, the original traceback is silently dropped.

Use exception chaining so the root cause is preserved:

```python
raise VulnReportTaskError(osv_data=osv_data) from osv_data
```

Without this, debugging production failures becomes significantly harder — you only see where `VulnReportTaskError` was raised, not the actual error from the background thread.

pulp_service/pulp_service/app/exceptions.py
```diff
@@ -0,0 +1,52 @@
...
+
+class UnsupportedPackageTypeError(PulpException):
+    """
+    Raised when a content type cannot be mapped to a supported package ecosystem.
+    """
+
+    error_code = "SVC0004"
+
+    def __str__(self):
+        return f"[{self.error_code}] " + _("Package type not supported!")
```
decko, line 52: Consider accepting the unsupported content/package type as an argument and including it in the message. Right now `[SVC0004] Package type not supported!` gives no hint about *which* type failed, making it hard to diagnose in production logs without correlating other output.

```python
def __init__(self, content_type=None):
    self.content_type = content_type
    super().__init__()

def __str__(self):
    if self.content_type:
        return f"[{self.error_code}] " + _("Package type not supported: {content_type}").format(content_type=self.content_type)
    return f"[{self.error_code}] " + _("Package type not supported!")
```

Minor nit: error messages conventionally don't use exclamation marks.

PR #1231, opened 2026-05-26
fix: allow authenticated users to read PyPI views and public domains

decko:
## Summary

- Move PyPI (`isinstance(view, PyPIMixin)`) and public-domain (`"public-" in domain.name`) permission checks **before** the authentication gate in `DomainBasedPermission.has_permission()`, so both anonymous and authenticated users get read access on safe methods
- Remove the dead `AllowUnauthPull` class (unreferenced since patch 0038 was deleted in 5096c48)

## Problem

After PR #1220 moved the PyPI anonymous read logic from patch 0038 into `DomainBasedPermission`, the check was placed inside the `if not user.is_authenticated` branch. Authenticated users (e.g. `calunga-automatic-builds` org:15322645, `konflux-ansible-tenant` org:1979710) hitting public PyPI endpoints were routed to `_has_domain_access()`, which returned 403 because their org has no `DomainOrg` entry for the `public-trusted-libraries` domain.

**Timeline**: Last successful authenticated request was May 21 13:08 UTC (image `1222d66`). First 403 was May 22 00:41 UTC (image `278a49d`).

## Test plan

- [ ] Authenticated user can GET `/api/pypi/public-trusted-libraries/main/pypi/<package>/json/` (was 403, should be 200)
- [ ] Authenticated user can GET `/api/pypi/public-trusted-libraries/main/simple/<package>/` (was 403, should be 200)
- [ ] Anonymous user can still GET the same endpoints (unchanged, already 200)
- [ ] Authenticated user on a private domain they don't own still gets 403 for non-PyPI views
- [ ] Authenticated user on their own domain still works as before
- [ ] POST/PUT/DELETE on PyPI views still requires domain access (SAFE_METHODS guard)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Adjust domain-based permission checks so safe-method access to PyPI views and public domains is granted regardless of authentication status, and remove an unused permission class.

Bug Fixes:
- Ensure authenticated users can perform safe-method requests against PyPI views and public domains that were previously incorrectly rejected with 403 responses.

Enhancements:
- Unify the handling of safe-method access for PyPI views and public domains by moving the checks outside the anonymous-user-only branch in the permission logic.

Chores:
- Remove the unused AllowUnauthPull permission class.

### DIFF
--- a/pulp_service/pulp_service/app/authorization.py
+++ b/pulp_service/pulp_service/app/authorization.py
@@ -48,16 +48,17 @@ class DomainBasedPermission(BasePermission):
 
         user = request.user
 
-        # Anonymous users can make safe requests on public domains and PyPI views
-        if not user.is_authenticated:
-            if request.method in SAFE_METHODS:
-                from pulp_python.app.pypi.views import PyPIMixin
+        # Allow safe requests on PyPI views and public domains for all users
+        if request.method in SAFE_METHODS:
+            from pulp_python.app.pypi.views import PyPIMixin
 
-                if isinstance(view, PyPIMixin):
-                    return True
-                domain = getattr(request, "pulp_domain", None)
-                if domain and "public-" in domain.name:
-                    return True
+            if isinstance(view, PyPIMixin):
+                return True
+            domain = getattr(request, "pulp_domain", None)
+            if domain and "public-" in domain.name:
+                return True
+
+        if not user.is_authenticated:
             return False
 
         # Check if user is creating a domain or creating a resource within a domain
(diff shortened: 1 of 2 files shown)


pulp_service/pulp_service/app/authorization.py
```diff
@@ -48,16 +48,17 @@ def has_permission(self, request, view):
 
         user = request.user
 
-        # Anonymous users can make safe requests on public domains and PyPI views
+        # Allow safe requests on PyPI views and public domains for all users
+        if request.method in SAFE_METHODS:
+            from pulp_python.app.pypi.views import PyPIMixin
```
decko: What if we move this import to the header, and then check if the view is a PyPIMixin instance to then run the other checks?

pulp_service/pulp_service/app/authorization.py
```diff
@@ -48,16 +48,17 @@ def has_permission(self, request, view):
...
+
+            if isinstance(view, PyPIMixin):
+                return True
+            domain = getattr(request, "pulp_domain", None)
+            if domain and "public-" in domain.name:
+                return True
+
         if not user.is_authenticated:
```
decko, line 61: What will happen with this check? How it was before the commit that changed it?

PR #1316, opened 2026-07-06
feat: gate PyPI access via distribution content guards instead of hardcoded feature check

CryptoRodeo:
DomainBasedPermission no longer hardcodes the "lightwell" domain name or the "lightwell-network" feature.

Instead, _check_pypi_safe_method_access reads the distribution's content_guard:
- if one is configured, it delegates to guard.cast().permit(request).
- if none, SAFE_METHOD access is open (preserving the existing default).

DomainOrg owners still bypass the guard.

This makes PyPI access control domain-name-agnostic, any distribution can be protected by attaching a content guard, with no code changes needed.

JIRA: PULP-1993

## Summary by Sourcery

Gate PyPI SAFE_METHOD access for content-guarded distributions via their content guard, while preserving default-allow behavior for unguarded distributions and DomainOrg-based bypass.

Enhancements:
- Replace hardcoded lightwell-network feature entitlement checks with distribution content-guard-driven permission checks for PyPI views.
- Simplify DomainBasedPermission by removing lightwell-specific feature constants and helper methods in favor of a generic content guard flow.
- Extend and update unit tests to cover content-guard-based PyPI access decisions, including DomainOrg bypass and domain-name agnostic behavior.

Tests:
- Refactor PyPI permission tests to construct views with mock content guards and verify guard permit() usage, denial paths, and DomainOrg bypass scenarios.

## Summary by Sourcery

Gate PyPI SAFE_METHOD access using distribution content guards instead of a hardcoded feature flag, while preserving default-allow behavior for unguarded distributions and DomainOrg-based bypass.

Enhancements:
- Refactor DomainBasedPermission to consult a distribution’s content guard for PyPI SAFE_METHOD access, replacing lightwell-specific feature checks and constants with a domain-agnostic flow.
- Improve robustness of PyPI permission handling by explicitly handling distribution resolution errors, content-guard casting failures, and unexpected permit() exceptions with fail-closed behavior and structured logging.
- Clarify and slightly reorder URL routing imports without changing behavior.

Tests:
- Expand unit tests for DomainBasedPermission to cover content-guard-driven PyPI access decisions, DomainOrg bypass, domain-name agnostic behavior, and error-handling/logging paths.
- Update functional tests to exercise content-guard-based PyPI access control, including feature-entitled and non-entitled orgs, domain-owner bypass, unauthenticated access, write-operation behavior, and unaffected non-PyPI endpoints.
- Adjust lightwell read-only group functional tests to use a content-guarded PyPI distribution within the lightwell domain.

### DIFF
--- a/pulp_service/pulp_service/app/authorization.py
+++ b/pulp_service/pulp_service/app/authorization.py
@@ -6,12 +6,13 @@ from contextvars import ContextVar
 
 import jq
 from django.db.models import Q
+from django.http import Http404
 from rest_framework.permissions import SAFE_METHODS, BasePermission
 
 from pulpcore.plugin.models import Domain
 from pulpcore.plugin.util import extract_pk, get_domain_pk
 
-from pulp_service.app.models import DomainOrg, FeatureContentGuard
+from pulp_service.app.models import DomainOrg
 
 _logger = logging.getLogger(__name__)
 org_id_var = ContextVar("org_id")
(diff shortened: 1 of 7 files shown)


dkliban:
## Architectural Review Findings

### M1. Bare `except Exception` on `view.distribution` is too broad

The `try/except Exception: return None` block around `view.distribution` swallows `Http404` (distribution not found) and signals "not a PyPI view, fall through to standard DomainOrg checks." This conflates "distribution doesn't exist" with "not a PyPI view." Database timeouts or connection errors would also silently fall through rather than failing closed.

**Suggestion**: Catch `Http404` specifically and let the view handle the 404 naturally. For unexpected exceptions, fail closed:

```python
from django.http import Http404

try:
    distribution = view.distribution
except Http404:
    return True  # let the view raise 404 normally
except Exception:
    _logger.exception("Unexpected error resolving distribution for PyPI permission check")
    return False  # fail closed
```

### M2. `guard.cast()` failure path is unhandled

`guard.cast()` performs a DB lookup (`self._pulp_model_map[self.pulp_type].objects.get(pk=self.pk)`). If the content guard subclass row has been deleted (orphaned FK, race condition with concurrent deletion), this raises `ObjectDoesNotExist` or `KeyError` — neither is caught by `except PermissionError`, so it would bubble up as an unhandled 500 instead of failing closed.

**Suggestion**: Separate the `cast()` call from `permit()` and handle each failure mode:

```python
try:
    casted_guard = guard.cast()
except Exception:
    _logger.exception("Failed to resolve content guard type for distribution")
    return False  # fail closed

try:
    casted_guard.permit(request)
    # ... log and return True
except PermissionError:
    # ... log and return False
```

### M3. Stale comments on `LIGHTWELL_DOMAIN_NAME` constant

The comment above `LIGHTWELL_DOMAIN_NAME = "lightwell"` still says PyPI views are gated by domain name and the lightwell-network feature. After this PR, that constant is only used for the readonly group logic. The comment should be updated to reflect this.

CryptoRodeo:
@sourcery-ai review

CryptoRodeo:
@sourcery-ai review

dkliban (approved):

PR #1325, opened 2026-07-15
feat: extend DomainBasedPermission to check lightwell subscription via Feature Service

carlosthe19916:
## Summary by Sourcery

Integrate subscription-based access checks for Lightwell domain content listing into DomainBasedPermission using the feature service, and add configuration plus functional coverage for this behavior.

New Features:
- Grant SAFE_METHOD content listing access in the Lightwell domain when the requesting org has the configured Lightwell feature via the feature service, independent of DomainOrg or read-only group membership.

Enhancements:
- Introduce a helper to detect Lightwell content listing requests and centralize feature-based subscription checking via FeatureContentGuard.
- Add a configurable constant for the Lightwell feature name used when querying the feature service.

Tests:
- Add functional tests exercising Lightwell content listing permissions against the real feature service for entitled, non-entitled, domain-owner, and unauthenticated scenarios.

### DIFF
--- a/pulp_service/pulp_service/app/authorization.py
+++ b/pulp_service/pulp_service/app/authorization.py
@@ -5,7 +5,6 @@ from binascii import Error as Base64DecodeError
 from contextvars import ContextVar
 
 import jq
-from django.conf import settings
 from django.db.models import Q
 from django.http import Http404
 from rest_framework.permissions import SAFE_METHODS, BasePermission
@@ -32,6 +31,9 @@ LIGHTWELL_DOMAIN_NAME = "lightwell"
 # NOT grant write access, and it does NOT apply to PyPI views, which are gated
 # by the distribution's content guard (see _check_pypi_safe_method_access()).
 LIGHTWELL_READONLY_GROUP_NAME = "Lightwell-ReadOnly"
+# Fixed identifier in the Features Service; checked by _check_lightwell_subscription() to
+# gate content listing access for orgs with an active lightwell subscription.
+LIGHTWELL_FEATURE_NAME = "lightwell-network"
 
 
 class DomainBasedPermission(BasePermission):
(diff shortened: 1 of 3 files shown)


dkliban:
## Architecture Review Feedback

### 1. Unhandled exceptions from Features Service (Major)

`_check_lightwell_subscription` only catches `PermissionError`, but `check_feature` makes an HTTP call that can raise `ConnectionError`, `Timeout`, or `RequestException`. During a Features Service outage, every content listing request to the lightwell domain would 500.

Suggest catching `Exception` broadly, logging it, and returning `None` so the request falls through to the existing `Lightwell-ReadOnly` group check below. This matches how `_check_pypi_safe_method_access` handles similar failures:

```python
except Exception:
    _logger.exception("Unexpected error evaluating content guard permit")
    return False
```

### 2. Use a module-level constant instead of a Django setting (Minor)

`LIGHTWELL_FEATURE_NAME` is added to `settings.py`, but the other lightwell-specific values are hardcoded as module-level constants in `authorization.py`:

```python
LIGHTWELL_DOMAIN_NAME = "lightwell"
LIGHTWELL_READONLY_GROUP_NAME = "Lightwell-ReadOnly"
```

The feature name `"lightwell-network"` is a fixed identifier in the Features Service — no more likely to change per deployment than the domain name or group name. A constant in `authorization.py` would be more consistent and avoids implying this is an operator-tunable knob:

```python
LIGHTWELL_FEATURE_NAME = "lightwell-network"
```

### 3. Path substring matching instead of view-type dispatch (Minor)

`_is_content_listing_request` uses a substring check on `PATH_INFO`:

```python
return "/api/v3/content/" in request.META.get("PATH_INFO", "")
```

The existing PyPI check uses view-type inspection (`isinstance(view, PyPIMixin)`), which is more robust against future URL routing changes. Consider checking via view class or view name instead, or at minimum add a comment explaining why substring matching was chosen here.

### 4. Test coverage gaps (Note)

The functional tests cover the primary cases well. A few additional scenarios would strengthen coverage:

- **Write method (POST) to content listing endpoint by subscribed user** — verifies the SAFE_METHODS gate survives refactoring
- **User in Lightwell-ReadOnly group but without subscription** — regression guard for the existing group-based path
- **Non-lightwell domain with `/api/v3/content/` path by subscribed user** — confirms the domain name guard prevents cross-domain access

carlosthe19916:
@dkliban I think I addressed your points

dkliban (approved):

PR #1337, opened 2026-07-16
feat: add request_time_ms to pulp-content access logs and exporter

CryptoRodeo:
Add request duration to the content access log format and export it as request_time_ms (int32, truncated) in Parquet. Parser handles old log lines without the field gracefully.

Uses aiohttp's %Tf atom (fractional seconds) instead of the originally planned %(M)s (gunicorn milliseconds) because pulp-content runs aiohttp.GunicornWebWorker, which explicitly rejects gunicorn-style %(...)s format patterns at startup.

Using %(M)s would crash the service. Since aiohttp has no native millisecond atom, the exporter converts seconds to integer milliseconds (float * 1000, truncated).

## Summary by Sourcery

Capture request duration from pulp-content access logs and export it as a request_time_ms column in Parquet while preserving compatibility with existing log formats.

New Features:
- Add request_time_ms as a derived request duration field in exported pulp-content Parquet data

Enhancements:
- Extend the content access log parser to capture optional request_time values while remaining backward compatible with older log lines

Build:
- Update pulp-content deployment configuration to emit request_time in access logs using aiohttp’s fractional-seconds format atom

Tests:
- Add unit tests covering request_time parsing, milliseconds conversion, inclusion in Python/RPM/Maven export paths, and handling of legacy logs without the field

### DIFF
--- a/management_tools/pulp-access-logs-exporter/src/pulp_access_logs_exporter/content_cloudwatch.py
+++ b/management_tools/pulp-access-logs-exporter/src/pulp_access_logs_exporter/content_cloudwatch.py
@@ -60,7 +60,7 @@ def _parse_request_time_ms(value):
     if not value or value == "-":
         return None
     try:
-        return int(float(value) * 1000)
+        return round(float(value) * 1000)
     except (ValueError, TypeError):
         return None
 
--- a/management_tools/pulp-access-logs-exporter/src/pulp_access_logs_exporter/content_cloudwatch.py
+++ b/management_tools/pulp-access-logs-exporter/src/pulp_access_logs_exporter/content_cloudwatch.py
@@ -56,6 +56,15 @@ def _parse_org_id(org_value):
     return org_value
 
 
+def _parse_request_time_ms(value):
+    if not value or value == "-":
+        return None
+    try:
+        return int(float(value) * 1000)
+    except (ValueError, TypeError):
+        return None
+
+
 def _parse_timestamp(timestamp_str):
     if not timestamp_str:
         return None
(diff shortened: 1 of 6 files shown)


management_tools/pulp-access-logs-exporter/src/pulp_access_logs_exporter/content_cloudwatch.py
```diff
@@ -56,6 +56,15 @@ def _parse_org_id(org_value):
     return org_value
 
 
+def _parse_request_time_ms(value):
+    if not value or value == "-":
+        return None
+    try:
+        return int(float(value) * 1000)
+    except (ValueError, TypeError):
```
decko, line 64: **Suggestion:** `int()` truncates, which introduces off-by-one errors for ~1% of exact-millisecond boundaries above 1 second due to IEEE 754 float representation (e.g. `"1.001000"` → 1000 instead of 1001). `round()` eliminates this:
```suggestion
        return round(float(value) * 1000)
```

decko (approved):
Clean implementation — backward compatibility, rolling deployment safety, regex correctness, and schema consistency all check out. One minor suggestion inline (see comment on `_parse_request_time_ms`).

management_tools/pulp-access-logs-exporter/src/pulp_access_logs_exporter/content_cloudwatch.py
CryptoRodeo: How's this? -> https://github.com/pulp/pulp-service/pull/1337/changes/8c7be2751a168d15ffa06f4da03a7ee4be77fc5d

decko:
The merge is yours @CryptoRodeo

PR #1351, opened 2026-07-22
Update patch 0018 to stop re-rooting pulp-container paths

git-hyagi:
ref: https://redhat.atlassian.net/browse/PULP-2092

## Summary by Sourcery

Enhancements:
- Adjust the existing registry API patch so that it no longer rewrites or re-roots pulp-container paths.


dkliban:
@git-hyagi i expected the patch file to get smaller. urls.py should not be part of the patch. what do you think?

git-hyagi:
hmm... maybe i am misunderstanding the task(s)
from https://redhat.atlassian.net/browse/PULP-2092

>  Both the API and content routes should be kept as they are upstream 

I thought we should also update urls.py.


And from (https://redhat.atlassian.net/browse/PULP-2093):

> Update the Akamai rules to route pulp-container traffic to the upstream paths:
>     /v2/ traffic should be routed to pulp-api
>     /extensions/v2/ traffic should be routed to pulp-api

without the urls.py change, these akamai rules would also not be needed, right?

dkliban (approved):

PR #1355, opened 2026-07-23
feat: Agentic SDLC Pipeline V1 — plan, elaborate, orchestrate, implement, review

decko:
## Summary

Implements the five-workflow event-driven SDLC pipeline connecting Jira planning to automated implementation.

### Workflows
- **W1 Planner** (evolved) — conversational planning with iterative refinement, produces SPEC + TASK OUTLINE + DEPENDENCIES format
- **W2 Task Elaborator** (new) — creates rich child task tickets from approved plans, triggered by `plan-approved` label
- **W3 Execution Orchestrator** (new) — labels next ready tasks after completion, detects all-done, handles failures, triggered by `executing` label
- **W4 SDLC Planned Pipeline** (new) — implements one planned task per run: detect plan → fetch spec → implement → verify → review → PR → merge
- **W5 Human Review** — handled by W3 creating a review task when all implementation tasks complete

### Agents
- **Planner** (evolved) — new plan format with SPEC, TASK OUTLINE, DEPENDENCIES sections; previous plan injection for re-planning
- **Task Elaborator** (new) — reads plan, creates Jira tasks with structured descriptions
- **Pulp Developer** (evolved) — guided mode for planned tasks, 30-min timeout
- **Go Specialist Reviewer** (new) — reviews Go code for idiomatic patterns and concurrency safety
- **Pulp Domain Reviewer** (new) — reviews for multi-tenancy, auth, upstream compatibility
- **Execution Orchestrator Agent** (new) — parses DEPENDENCIES block, manages task labeling

### Dependencies
- alcove v0.55.0 deployed (alcove-ai/alcove#773) ✅
- `git-ssl-no-verify` credential on Hosted Pulp team ✅

### Design Doc
Full design at `docs/plan-format-design.md`.

## Test Plan
- [ ] Sync agents, trigger W1 on a test epic with `needs-planning`
- [ ] Comment on epic to test conversational re-planning
- [ ] Add `plan-approved` label to test W2 task creation
- [ ] Verify W4 picks up a ready task with `agentic-sdlc`
- [ ] Verify W3 labels next task after completion
- [ ] Test failure handling: plan deviation, CI failure, review rejection

Assisted-by: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

## Summary by Sourcery

Introduce an agentic, plan-driven SDLC pipeline for pulp-service that connects Jira epic planning to automated task creation, guided implementation, review, and orchestration.

New Features:
[...]

PR #1357, opened 2026-07-23
fix: replace null conditions with total-based checks for condition evaluator

decko:
## Summary
The alcove condition evaluator does not support `== null` / `!= null` syntax — it only handles single-quoted strings and numeric comparisons. All four SDLC pipeline workflows fail to parse on sync.

Replaces all `issue_keys == null` / `issue_keys != null` conditions with `total == 0` / `total > 0`. Adds `total` to search step outputs where missing.

## Note
`total` may be unreliable when results exist (alcove-ai/alcove#772) — it can return 0 even with matches. But it correctly returns 0 for the no-results case, which is what these conditions check. Filing a separate alcove issue for null comparison support in the condition evaluator.

## Test plan
- [ ] Sync agents — verify all 4 workflows parse without errors
- [ ] Trigger planner on an epic — verify `create_plan_task` condition works

Assisted-by: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

## Summary by Sourcery

Align SDLC-related workflows with the condition evaluator by replacing null-based conditions with total-based checks and exposing total in relevant search step outputs.

Bug Fixes:
- Ensure SDLC planned workflow conditions correctly detect absence or presence of related plan issues using total-based checks instead of null comparisons.
- Fix execution-orchestrator workflow logic that depends on plan task detection by switching issue existence checks from null comparisons to total-based conditions.
- Correct planner workflow conditions for creating or fetching implementation plan tasks to rely on total counts rather than unsupported null equality checks.
- Update task-elaborator workflow conditions to use total-based checks for detecting missing or existing implementation plan tasks, preventing condition evaluation failures.

Enhancements:
- Add total to jira-search-issues step outputs across SDLC workflows to support reliable issue count-based conditional logic.

### DIFF
--- a/.alcove/workflows/execution-orchestrator.yml
+++ b/.alcove/workflows/execution-orchestrator.yml
@@ -32,7 +32,7 @@ workflow:
     inputs:
       jql: "parent = {{trigger.issue_key}} AND summary ~ 'Implementation Plan'"
       max_results: 1
-    outputs: [issue_keys]
+    outputs: [issue_keys, total]
 
   # Step 4: Find all child tasks (excluding the plan task)
   - id: get_children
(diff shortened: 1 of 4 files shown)


decko:
Not needed — the conditions on main already use == null (from PR #1355), and alcove v0.56.0 now supports null comparisons. The only remaining diff was a comment update.

PR #1368, opened 2026-07-28
Add If-Modified-Since header support to content app

git-hyagi:
Patch pulpcore content handler to return Last-Modified headers and honor If-Modified-Since conditional requests, enabling clients to skip re-downloading unchanged content artifacts.

ref: https://redhat.atlassian.net/browse/PULP-2162

## Summary by Sourcery

Add support for conditional HTTP content handling by applying a new pulpcore patch in the Docker image.

Enhancements:
- Enable content responses to include Last-Modified headers and respect If-Modified-Since requests to avoid unnecessary artifact downloads.

Build:
- Update the Docker image build to apply a new pulpcore patch for If-Modified-Since header handling.

### DIFF
--- a/images/assets/patches/0062-Add-if-modified-since-header-support.patch
+++ b/images/assets/patches/0062-Add-if-modified-since-header-support.patch
@@ -1,0 +1,133 @@
+diff --git a/pulpcore/cache/cache.py b/pulpcore/cache/cache.py
+index ec6ca80e6..d98358718 100644
+--- a/pulpcore/cache/cache.py
++++ b/pulpcore/cache/cache.py
+@@ -4,7 +4,7 @@ import time
+ from functools import wraps
+ 
+ from aiohttp.web import FileResponse, HTTPSuccessful, Request, Response, StreamResponse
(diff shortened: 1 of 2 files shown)


images/assets/patches/0062-Add-if-modified-since-header-support.patch
```diff
@@ -0,0 +1,133 @@
...
+ from pulpcore.responses import ArtifactResponse
+@@ -356,13 +358,33 @@ class AsyncContentCache(AsyncCache):
+                 response = await self.make_entry(
+                     key, bk, func, args, kwargs, self.default_expires_ttl
+                 )
+-            elif size := response.headers.get("X-PULP-ARTIFACT-SIZE"):
+-                artifacts_size_counter.add(size)
++            else:
++                self._check_not_modified(request, response)
++                if size := response.headers.get("X-PULP-ARTIFACT-SIZE"):
```
dkliban, line 30: should we swap these two lines around? so even if we return a 304 the counter is updated?

PR #1375, opened 2026-07-29
Add If-Modified-Since header support to content app

git-hyagi:
Patch pulpcore content handler to return Last-Modified headers and honor If-Modified-Since conditional requests, enabling clients to skip re-downloading unchanged content artifacts.
ref: https://redhat.atlassian.net/browse/PULP-2162

## Summary by Sourcery

Add a new pulpcore patch to enable Last-Modified/If-Modified-Since HTTP header support in the content app and wire it into the Docker image build.

New Features:
- Support conditional GETs in the content app using Last-Modified and If-Modified-Since headers.

Build:
- Include and apply a new pulpcore patch file in the Dockerfile during image build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -156,6 +156,9 @@ RUN patch -p1 -d /usr/local/lib/pulp/lib/python${PYTHON_VERSION}/site-packages <
 COPY images/assets/patches/0061-Release-locks-on-fetch_task-exception.patch /tmp/
 RUN patch -p1 -d /usr/local/lib/pulp/lib/python${PYTHON_VERSION}/site-packages < /tmp/0061-Release-locks-on-fetch_task-exception.patch
 
+COPY images/assets/patches/0062-Add-if-modified-since-header-support.patch /tmp/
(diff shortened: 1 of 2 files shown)


git-hyagi:
The tests are failing because of a missing fixture: rpm_package_api.
From claude:
```
  Root cause: In pulp_rpm 3.38.2 (pinned in requirements.txt), the rpm_repository_factory fixture in pytest_plugin.py depends on rpm_package_api, but that fixture was defined in                                   
  pulp_rpm/tests/functional/conftest.py — not in pytest_plugin.py. When pulp_rpm's own tests run, both files are loaded so it works fine. But when pulp_service's tests run, only pytest_plugin.py is loaded (via   
  the pytest11 entry point), so rpm_package_api is missing.                                                                                                                                                         
                                                                                                                                                                                                                    
  This was fixed upstream in commit 7028a2e61 ("Move rpm_package_api fixture to pytest_plugin"), but that hasn't been released yet — it's after the 3.38.2 tag.                                                     
                                                                                                                                                                                                                    
  Fix options:                                                                                                                                                                                                      
                                                                                                                                                                                                                    
  1. Add rpm_package_api fixture to pulp_service's conftest.py — quick workaround until the next pulp_rpm release                                                                                                   
  2. Pin pulp_rpm to a git ref that includes the fix — couples you to an unreleased version
 ```

